PR #23, opened 2018-02-18
fix data race between .Run() and .Stop(). 

jason-xxl:
It (at least) affects some test case that open and close server fast.


jason-xxl:
https://github.com/carousell/Orion/issues/74
think they are just same issue.

PR #28, opened 2018-04-30
ordered svc startup

rajathishere:
* capture service name on first register in order
* reload services in order on signal


ankurs:
this is no handled by ServiceFactoryV2

PR #42, opened 2018-05-25
adding support for options from ORION:OPTION: <option>

ankurs:
fixing #33


orion/core.go
```diff
@@ -132,6 +140,17 @@ func (d *DefaultServerImpl) AddDefaultDecoder(serviceName string, decoder Decode
 	d.defDecoders[serviceName] = decoder
 }
 
+func (d *DefaultServerImpl) AddOption(serviceName, method, option string) {
+	if d.options == nil {
+		d.options = make(map[string]*optionInfo)
+	}
+	d.options[serviceName+":"+method] = &optionInfo{
```
vvarma: fmt is cleaner?

PR #73, opened 2018-08-08
Adding support for Websockets to gRPC streams

ankurs:


orion/handlers/http/defaults.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func deserialize(ctx context.Context, data []byte, r interface{}) error {
+	serType := ContentTypeFromHeaders(ctx)
+	switch serType {
+	case modifiers.ProtoBuf:
+		if protoReq, ok := r.(proto.Message); ok {
+			return proto.Unmarshal(data, protoReq)
+		}
+		fallthrough
```
vvarma: bit unsafe, if content type is for proto then if there isnt a proto message in the request, should fail?

orion/handlers/http/handler.go
```diff
@@ -0,0 +1,207 @@
...
+func (h *httpHandler) AddMiddleware(serviceName string, method string, middlewares ...string) {
+	if h.middlewares == nil {
+		h.middlewares = handlers.NewMiddlewareMapping()
+	}
+	h.middlewares.AddMiddleware(serviceName, method, middlewares...)
+}
+
+func (h *httpHandler) Run(httpListener net.Listener) error {
+	r := mux.NewRouter()
+	fmt.Println("Mapped URLs: ")
```
vvarma: fmt?

orion/handlers/http/defaults.go
ankurs: yes the function should fail and return the error, but in case of get requests we check for io.EOF and don't call this function

PR #75, opened 2018-09-04
adding grpc error type support in errors

ankurs:


utils/errors/errors.go
```diff
@@ -70,6 +76,13 @@ func (c customError) Cause() error {
 	return c.cause
 }
 
+func (c customError) GRPCStatus() *status.Status {
+	if c.status == nil {
+		return status.New(codes.Unknown, "error occured")
```
hothero: do we can use c.Msg to replace the fixed string?

ankurs: I this is better if i add Error(), will do that!

PR #86, opened 2018-10-24
Fix go routine pile up

ankurs:


utils/listenerutils/listenerutils.go
```diff
@@ -34,19 +38,82 @@ func (c *customListener) Close() error {
...
+		}
 		go func() {
 			// wait for timeout after stop and close all active connections
-			<-c.stop
-			time.Sleep(c.timeout)
-			connection.conn.Close()
+			select {
+			case <-c.stop:
+				time.Sleep(c.timeout)
+				connection.conn.Close()
```
vvarma: conn.Close() instead?

ankurs: yes, good catch! 
let me update!

PR #95, opened 2019-03-28
Error supports raw data

hothero:
sample usage: https://github.com/carousell/listing/pull/286
this PR can also tackle for the case when you want to log value from request rather than reporting dynamic error message: https://github.com/carousell/offer/commit/12d647cb5442cd2e6747242620664d7d35031167

test snippets
```
e := errors.New("test raw data")
notifier.Notify(errors.WrapWithRawData(e, "", req)
```
image


utils/errors/errors.go
```diff
@@ -145,26 +155,31 @@ func NewWithSkip(msg string, skip int) ErrorExt {
...
 //WrapWithSkip wraps an existing error and appends stack information if it does not exists skipping the number of function on the stack
 func WrapWithSkip(err error, msg string, skip int) ErrorExt {
-	return WrapWithSkipAndStatus(err, msg, skip+1, nil)
+	return wrapWithSkipAndStatus(err, msg, skip+1, nil)
 }
 
-//WrapWithSkip wraps an existing error and appends stack information if it does not exists skipping the number of function on the stack along with GRPC status
-func WrapWithSkipAndStatus(err error, msg string, skip int, status *grpcstatus.Status) ErrorExt {
+//wrapWithSkipAndStatus wraps an existing error and appends stack information if it does not exists skipping the number of function on the stack along with GRPC status
+func wrapWithSkipAndStatus(err error, msg string, skip int, status *grpcstatus.Status, rawData ...interface{}) ErrorExt {
```
hothero: don't expose this real logic func

utils/errors/errors.go
```diff
@@ -145,26 +155,31 @@ func NewWithSkip(msg string, skip int) ErrorExt {
...
 }
 
 //Wrap wraps an existing error and appends stack information if it does not exists
 func Wrap(err error, msg string) ErrorExt {
 	return WrapWithSkip(err, msg, 1)
 }
 
-//Wrap wraps an existing error and appends stack information if it does not exists along with GRPC status
+//WrapWithRawData wraps an existing error and appends stack information if it does not exists
+func WrapWithRawData(err error, msg string, rawData ...interface{}) ErrorExt {
```
hothero: why create new func only for raw data instead of adding for each Wrap* functions? rawData is accepting all types of args, what if we need to add new args as interface....for example,

```
func Wrap(err error, msg string, rawData ...interface{}) ..
// become
func Wrap(err error, msg string, anything interface{}, rawData ...interface{})
```

in case of it, I would prefer create new function to prevent potential breaking change in the future.

utils/errors/errors.go
hothero: I've checked no one using it

changchingchen: Users may defined their own errors on top of Orion. If these function are not exposed, then the call stacks for functions that generate the error will not be skipped. Since one can only call Wrap function but it always skips starting from Wrap but not the wrapping function users defined.

hothero: sure, it make sense to me.

utils/errors/errors.go
```diff
@@ -37,13 +37,19 @@ type NotifyExt interface {
 	Notified(status bool)
 }
 
+// RawExt is the interface definition for raw data about an error
```
changchingchen: Is it necessary to create a new interface just  to implement this method or maybe it could be declared in the original ErrorExt interface? In my opinion, the more interface declared, the more effort that developers need to do to customize their own error struct on top of Orion error.

utils/errors/errors.go
```diff
@@ -129,7 +139,7 @@ func packageFuncName(pc uintptr) (string, string) {
 }
 
 //New creates a new error with stack information
-func New(msg string) ErrorExt {
+func New(msg string, rawData ...interface{}) ErrorExt {
```
benjaminheng: I'm against changing the function signature for `New()`. Traditionally `.New()` accepts only a single string.

utils/errors/errors.go
```diff
@@ -163,8 +178,8 @@ func WrapWithSkip(err error, msg string, skip int) ErrorExt {
 	return WrapWithSkipAndStatus(err, msg, skip+1, nil)
 }
 
-//WrapWithSkip wraps an existing error and appends stack information if it does not exists skipping the number of function on the stack along with GRPC status
-func WrapWithSkipAndStatus(err error, msg string, skip int, status *grpcstatus.Status) ErrorExt {
+//WrapWithSkipAndStatus wraps an existing error and appends stack information if it does not exists skipping the number of function on the stack along with GRPC status
```
benjaminheng: Update the godoc comment to describe the new rawdata argument.

utils/errors/errors.go
benjaminheng: By this reasoning we should not be modifying the `.New()` function either.

utils/errors/errors.go
benjaminheng: Actually, based on a previous comment:

> 
> 
> why create new func only for raw data instead of adding for each Wrap* functions? rawData is accepting all types of args, what if we need to add new args as interface....for example,
> 
> ```
> func Wrap(err error, msg string, rawData ...interface{}) ..
> // become
> func Wrap(err error, msg string, anything interface{}, rawData ...interface{})
> ```
> 
> in case of it, I would prefer create new function to prevent potential breaking change in the future.

This has the same issue of potentially introducing a breaking change in the future.

hothero: that's sort of true...I was thinking more about the purpose of function, so if the function indicating `RawData` it also implies you need to take the potential effort in the future for breaking change. Recently I have an idea on Notify func, will do the proposal soon to help the changes in this PR better.

utils/errors/errors.go
jettbow: Also rawData is unused in this function?

jettbow:
I am thinking we can use Functional Options pattern here.
Here is the reference http://tmrts.com/go-patterns/idiom/functional-options.html.
And here is the patch i created 
https://github.com/carousell/Orion/compare/master...error-supports-raw-data-patch

utils/errors/errors.go
hothero: I don't think more interfaces will make relative efforts to customizing on your own errors, thus it should make things easier. Ideally, you can embed interfaces you need rather than you are forced to implement all of functions if we put all functions in only one interface here.

hothero:
@jettbow looks a good idea! However, I rather prefer involving one main change than covering multiple proposals in a PR. Maybe you can create another PR to propose it!

jettbow:
@hothero Yes here it is. 
https://github.com/carousell/Orion/pull/115

utils/errors/errors.go
```diff
@@ -83,6 +90,14 @@ func (c customError) GRPCStatus() *grpcstatus.Status {
 	return c.status
 }
 
+func (c customError) RawData() []interface{} {
```
hrs7991: can make this a pointer receiver similar to `AppendRawData`

utils/errors/notifier/notifier.go
```diff
@@ -145,6 +145,10 @@ func NotifyWithLevelAndSkip(err error, skip int, level string, rawData ...interf
 		}
 		n.Notified(true)
 	}
+
+	if r, ok := err.(errors.RawExt); ok {
+		rawData = append(r.RawData(), rawData...)
```
hrs7991: any specific reason why have put our custom raw data ahead of the other data? basically comparing this vs `rawData = append(rawData, r.RawData())`

utils/errors/errors.go
hrs7991: I think we should add another `NewWithRawData` function as well.. thoughts?

hothero: it needs another func `WrapWithSkipAndRawData`, I have no strong opinion on this, for now I feel it's good enough because in most of cases we wrap data for returned error rather than creating an new error with raw data

utils/errors/notifier/notifier.go
hothero: no specific reason, just append it lol

achichen:
@hothero @benjaminheng could you take this old PR? if not relevant any more, we can close it

hothero:
we're also defining a new logging/notifier package and we have key/value arguments for reporting already.

PR #99, opened 2019-05-05
Patch http default deserialize to better handle error cases

jason-xxl:
> func deserialize(ctx context.Context, data []byte, r interface{}) error {
> 	serType := strings.TrimSpace(ContentTypeFromHeaders(ctx))
> 	if strings.Index(serType, ";") > -1 {
> 		// patch 1: to prevent cases like "application/json; charset=utf-8"
> 		serType = strings.TrimSpace(strings.Split(serType, ";")[0])
> 	}
> 
> 	switch serType {
> 	case modifiers.ProtoBuf:
> 		if protoReq, ok := r.(proto.Message); ok {
> 			return proto.Unmarshal(data, protoReq)
> 		}
> 		fallthrough
> 	case modifiers.JSONPB, modifiers.JSON, "":
> 		// path 2: "" for cases content-type is not passed properly
> 		if protoReq, ok := r.(proto.Message); ok {
> 			return jsonpb.UnmarshalString(string(data), protoReq)
> 		}
> 		fallthrough
> 	default:
> 		return json.Unmarshal(data, r)
> 	}
> }

### DIFF
--- a/orion/handlers/http/defaults.go
+++ b/orion/handlers/http/defaults.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"time"
+	"strings"
 
 	"github.com/carousell/Orion/orion/modifiers"
 	"github.com/golang/protobuf/jsonpb"
--- a/orion/handlers/http/defaults.go
+++ b/orion/handlers/http/defaults.go
@@ -41,14 +41,20 @@ func DefaultEncoder(req *http.Request, r interface{}) error {
 }
 
 func deserialize(ctx context.Context, data []byte, r interface{}) error {
-	serType := ContentTypeFromHeaders(ctx)
+	serType := strings.TrimSpace(ContentTypeFromHeaders(ctx))
+	if strings.Index(serType, ";") > -1 {
+		// patch 1: to prevent cases like "application/json; charset=utf-8"
+		serType = strings.TrimSpace(strings.Split(serType, ";")[0])
+	}
+
 	switch serType {
 	case modifiers.ProtoBuf:
 		if protoReq, ok := r.(proto.Message); ok {
 			return proto.Unmarshal(data, protoReq)
 		}
 		fallthrough
-	case modifiers.JSONPB:
+	case modifiers.JSONPB, modifiers.JSON, "":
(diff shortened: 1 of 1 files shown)


orion/handlers/http/defaults.go
```diff
@@ -41,14 +42,20 @@ func DefaultEncoder(req *http.Request, r interface{}) error {
...
+	}
+
 	switch serType {
 	case modifiers.ProtoBuf:
 		if protoReq, ok := r.(proto.Message); ok {
 			return proto.Unmarshal(data, protoReq)
 		}
 		fallthrough
-	case modifiers.JSONPB:
+	case modifiers.JSONPB, modifiers.JSON, "":
```
hothero: previously modifiers.JSON and "" will go to default section, what's the intent to make jsonpb as default?

jason-xxl: We have seen issue in client handling serialization with Enum, basically it has to treat enum as number instead of string like jsonpb does, which downgrades readability and potentially other aspects. Since we are already using gRPC/Protobuf and Protobuf proactively maintains this jsonpb around std json, I think the proper default fallback should be jsonpb. 

I understand there's certain perf degrade in jsonpb as it's built on top of json, but I believe that should be trivial in whole request handling process. bottleneck is more at on storage level or remote dep, not a local serialiser.

Also understand there'd be other services started with std json as default so may worry about such assumption level change. Alternatively I can work to make this defaults configurable, if that works.

orion/handlers/http/defaults.go
```diff
@@ -41,14 +42,20 @@ func DefaultEncoder(req *http.Request, r interface{}) error {
...
+		// path 2: "" for cases content-type is not passed properly
```
hothero: typo: patch

orion/handlers/http/defaults.go
hothero: as per discussion in slack, jsonpb unmarshal will break if the client passes a JSON string serialized by JSON marshaller and thus we go patch1 only first. Later incrementally rollout fallbacking to jsonpb unmarshaller logic

achichen:
Closing legacy inactive PR

PR #100, opened 2019-05-07
add support for redis brokers

languantan:
Add support for redis backends

## Changelog
- [x] add redis config
- [x] set default values for redis

### DIFF
--- a/utils/worker/config.go
+++ b/utils/worker/config.go
@@ -11,7 +11,7 @@ func getServerName(config Config) string {
 }
 
 func getRedisServerName(config Config) string {
-	return fmt.Sprintf("redis://%s%s%s%s",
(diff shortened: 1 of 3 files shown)


hothero:
@rajathishere @edwardsamuel can you please help review this change to support worker using Redis as datastore? not sure if you have concerns on system side 🙏

edwardsamuel:
After looking at the internal implementation of the library (https://github.com/RichardKnop/machinery/blob/07418869e268a380aea5b9aefbb0b9f5bb715a12/v1/brokers/redis/redis.go#L288), there is no message acknowledgment.
In case of failure/server restart, we might lose ongoing messages.

hothero:
close this PR at the moment until Redis in machinery has ack feature.

PR #102, opened 2019-06-10
Option to make JSONPB the default JSON encoder/decoder

humjosh:
For several services that use JSON, we would like to use the `jsonpb` package instead of the `json` package for request and response encoding. Clients should be able to pass `application/json` as the `Content-Type`/`Accept` header, and should not be required to pass `application/jsonpb`.

Currently you can force an endpoint to serialize out JSONPB with the `modifiers.SerializeOutJSONPB` modifier, but this is not an ideal solution because it prevents clients from specifying a different response format in the `Accept` header, such as protobuf.

In this PR I have added a new option for Orion servers: `DefaultJSONPB`. If set to `true`, any requests or responses which have specified `application/json` will use the `jsonpb` marshaller instead of the `json` marshaller. The default value for this setting is `false`, for backwards compatibility.

### DIFF
--- a/orion/handlers/http/defaults.go
+++ b/orion/handlers/http/defaults.go
@@ -43,17 +43,26 @@ func DefaultEncoder(req *http.Request, r interface{}, JSONPB bool) error {
 
 func deserialize(ctx context.Context, data []byte, r interface{}, JSONPB bool) error {
 	serType := ContentTypeFromHeaders(ctx)
-	if serType == modifiers.ProtoBuf {
+	switch serType {
+	case modifiers.ProtoBuf:
 		if protoReq, ok := r.(proto.Message); ok {
 			return proto.Unmarshal(data, protoReq)
 		}
-	}
-	if serType == modifiers.JSONPB || JSONPB {
+		fallthrough
+	case modifiers.JSONPB:
 		if protoReq, ok := r.(proto.Message); ok {
 			return jsonpb.UnmarshalString(string(data), protoReq)
 		}
+		fallthrough
+	default:
+		// if server preference is JSONPB, JSONPB should be used instead of JSON for unmarshalling
+		if JSONPB {
+			if protoReq, ok := r.(proto.Message); ok {
+				return jsonpb.UnmarshalString(string(data), protoReq)
+			}
+		}
+		return json.Unmarshal(data, r)
 	}
-	return json.Unmarshal(data, r)
 }
 
 // DefaultWSUpgrader upgrades a websocket if none are registered.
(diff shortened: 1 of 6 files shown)


orion/handlers/http/http.go
```diff
@@ -204,20 +204,15 @@ func (h *httpHandler) serialize(ctx context.Context, msg proto.Message) ([]byte,
...
+	if serType == modifiers.ProtoBuf {
 		data, err := proto.Marshal(msg)
 		return data, ContentTypeProto, err
-	case modifiers.JSON:
-		fallthrough
-	default:
-		// modifiers.JSON goes in here
-		data, err := json.Marshal(msg)
-		return data, ContentTypeJSON, err
+	} else if serType == modifiers.JSONPB || h.config.DefaultJSONPB {
```
benjaminheng: I think there's a disparity in behavior here. If the default is `jsonpb`, and in my handler I set `modifiers.SerializeOutJSON(ctx)`, then I expect my response to be serialized using the `json` package, not `jsonpb`.

https://github.com/carousell/Orion/blob/f006d565d9ed9d1d28b35f702affc1faade479d0/orion/handlers/http/http.go#L193-L206

You can replace L203 with

```
if h.config.DefaultJSONPB {
  serType = modifiers.JSONPB
} else {
  serType = modifiers.JSON
}
```

then I don't think you need to change this switch-case block.

orion/handlers/http/defaults.go
```diff
@@ -37,25 +38,22 @@ func DefaultEncoder(req *http.Request, r interface{}) error {
...
-	switch serType {
-	case modifiers.ProtoBuf:
+	if serType == modifiers.ProtoBuf {
 		if protoReq, ok := r.(proto.Message); ok {
 			return proto.Unmarshal(data, protoReq)
 		}
-		fallthrough
-	case modifiers.JSONPB:
+	}
+	if serType == modifiers.JSONPB || JSONPB {
```
benjaminheng: Let's make the logic here more explicit, in case we choose to add new serialization types in the future.

Instead of `content-type == "application/jsonpb" || defaultJSONPB`, 
change to 
`content-type == "application/jsonpb" || (content-type == "application/json" && defaultJSONPB)`

orion/handlers/http/http.go
humjosh: yes good point - actually we also want to set `serType = modifiers.JSONPB` in the case that the `Accept` or `Content-Type` header is set to `"application/json"`. I've updated the code with something similar to your suggestion

PR #104, opened 2019-06-29
Add GRPCRecovery Interceptor

stevenCarousell:
* main purpose is to prevent silent crash because of server panic, should recover and
  send the error out (through notifier/log).
* add GRPCRecoveryInterceptor to DefaultInterceptors()
* in GRPCRecoveryInterceptor, it will use grpc_recovery middleware with defined handler
  return an **Orion/utils/errors ErrorExt**. The error will skip 4 stack frames and show the function
  triggered the panic.

**Concept Flow**
not include the change for http middlewares
[https://drive.google.com/file/d/13AkM-68szUG_ITRaWJUpnJGcQF4Xqb3J/view?usp=sharing](url)

**Panic Log with no skip** 

```
{"@timestamp":"2019-07-01T16:19:33.3244498Z","caller":"interceptors/interceptors.go:137","err":"runtime error: index out of range","grpcMethod":"/{service}/{rpcName}","level":"error",
"stack":[
{"file":"/go/src/github.com/.../vendor/github.com/carousell/Orion/interceptors/interceptors.go","line":153,"function":"GRPCRecoveryInterceptor.func1.1"},
{"file":"/usr/local/go/src/runtime/asm_amd64.s","line":522,"function":"call32"},
{"file":"/usr/local/go/src/runtime/panic.go","line":513,"function":"gopanic"},
{"file":"/usr/local/go/src/runtime/panic.go","line":44,"function":"panicindex"},
{"file":"/go/src/github.com/.../service.go","line":442,"function":"{panic func}"},
...
```
so skip 4 stack frames to remove non-related information.

**Example on Sentry (Internal)**
https://sentry.carou.sl/carousell/home/issues/244262/?query=is:unresolved


interceptors/interceptors.go
```diff
@@ -137,6 +139,16 @@ func ServerErrorInterceptor() grpc.UnaryServerInterceptor {
 	}
 }
 
+//GRPCRecoveryInterceptor recover if server panic, return error with panic stack
+func GRPCRecoveryInterceptor() grpc.UnaryServerInterceptor {
+	opts := []grpc_recovery.Option{
+		grpc_recovery.WithRecoveryHandler(func(p interface{}) (err error) {
```
hothero: hi, can you help try if we can move https://github.com/carousell/Orion/blob/master/orion/handlers/http/http.go#L39 to here?

hothero: ^ @ankurs can you take a look?

interceptors/interceptors.go
```diff
@@ -137,6 +139,16 @@ func ServerErrorInterceptor() grpc.UnaryServerInterceptor {
...
+			return errors.NewWithSkip(fmt.Sprintf("%v", p), 7)
```
benjaminheng: What's the significance of these 7 frames?

benjaminheng:
Out of curiosity, in what scenario are you facing a panic from a GRPC handler?

hothero:
@benjaminheng the panic is from service call, so currently our Orion server will crash if service call happens panic on GRPC request then restart. But it won't happen on HTTP request.

interceptors/interceptors.go
stevenCarousell: it will list the line of wrap error,
UnaryServerInterceptor layer, 
then to the specific line of panic.
It seems we can skip the non related stack frames and directly show the panic line.

stevenCarousell:
@benjaminheng when the server handle the grpc request, if any panic in the handler, the server will crash without any recover mechanism. 
Cases like 1. index out of bound 2. concurrent map read/write 3. non-safe type assertion 4. etc.
If we add the recover mechanism, we can minimize the impact of panic and got the notify on sentry.

interceptors/interceptors.go
stevenCarousell: like gin skip 3 stacks which is non-related to the panic.
https://github.com/gin-gonic/gin/blob/master/recovery.go#L54

benjaminheng:
Oh, this is for incoming requests. Was under the impression that this was for outgoing ones

interceptors/interceptors.go
stevenCarousell: got it, will check the http part.

stevenCarousell: Hi @hothero, I've checked the content of http handler part

getHTTPHandler
-> httpHandler: handle new relic, add log (k, v) to context, panic recover, notifier
	-> serveHTTP
		-> 1. handle usual http handler, set by 
			`func (h *httpHandler) AddHTTPHandler(serviceName string, method string, path string, handler handlers.HTTPHandler)`
		   2. http handler carried with info.svc.svc
		   3. grpc converted to http request, only this type will apply interceptors we defined in DefaultInterceptors()

If we remove the implementation
https://github.com/carousell/Orion/blob/master/orion/handlers/http/http.go#L39 
for case 1, 2, there will be no protection, monitoring on them, only case 3 we will cover because of the interceptors.

I think the better way is for GRPCRecoveryInterceptor(), it should let httpHandler to protect itself and add the check to ignore http request.
`
if modifiers.IsHTTPRequest(ctx) {
			return handler(ctx, req)
		}
`

hothero: Could you help test this branch? https://github.com/carousell/Orion/pull/105 If it covers those cases you memtioned

stevenCarousell: ok, will check

PR #108, opened 2019-07-15
Add Unwrap method to customError

humjosh:
This will make `customError` compatible with the upcoming additions to the `errors` package in Go 1.13: `errors.Is` and `errors.As`. See [here](https://go.googlesource.com/proposal/+/master/design/go2draft-error-inspection.md) for details of the proposal, and see [here](https://godoc.org/golang.org/x/xerrors) for the existing implementation of this proposal, which will be merged into the standard library for Go 1.13.


ankurs:
looks good!

PR #112, opened 2019-07-26
Fix race condition in spanutils

itsneo:


benjaminheng:
Can you provide a description of how the race condition occurs and how this resolves it?

avence12:
Refer to - https://github.com/carousell/go-utils/blob/master/utils/spanutils/spanutils.go#L151

benjaminheng:
This repo is public and that isn't, I'm asking for the benefit of people outside of Carousell looking at this repository. To quote from the internal PR:

> A race condition occurs when GRPCTracingSpan is called from two separate goroutines, which might happen if I am making parallel gRPC calls to two different services. The problem is that we try to modify the metadata directly (which results in a panic if another goroutine tries to read it at the same time) - the fix is to make a copy and modify that instead.

jettbow:
Our team faced the same issue quoted below. 
From https://godoc.org/google.golang.org/grpc/metadata, the returned MD from FromIncomingContext should not be modified directly because it will cause some concurrency problems like `fatal error: concurrent map iteration and map write`. 

> This repo is public and that isn't, I'm asking for the benefit of people outside of Carousell looking at this repository. To quote from the internal PR:
> 
> > A race condition occurs when GRPCTracingSpan is called from two separate goroutines, which might happen if I am making parallel gRPC calls to two different services. The problem is that we try to modify the metadata directly (which results in a panic if another goroutine tries to read it at the same time) - the fix is to make a copy and modify that instead.

itsneo:
Here, we write values to the metadata from the context directly https://github.com/carousell/Orion/blob/1847293/utils/spanutils/spanutils.go#L217.

When another go-routine uses the same context to do the gRPC call, there is a chance we iterate (or read) the same metadata at the same time causing `fatal error: concurrent map iteration and map write`
https://github.com/grpc/grpc-go/blob/120728e/internal/transport/http2_client.go#L444.

PR #115, opened 2019-08-12
Error supports raw data by functional options pattern

jettbow:
This is another proposal for https://github.com/carousell/Orion/pull/95
Inspired by http://tmrts.com/go-patterns/idiom/functional-options.html.

### DIFF
--- a/utils/errors/errors.go
+++ b/utils/errors/errors.go
@@ -139,7 +139,7 @@ func packageFuncName(pc uintptr) (string, string) {
 }
 
 //New creates a new error with stack information
-func New(msg string) ErrorExt {
+func New(msg string, rawData ...interface{}) ErrorExt {
 	return NewWithSkip(msg, 1)
 }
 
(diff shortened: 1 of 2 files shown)


utils/errors/errors.go
```diff
@@ -37,13 +37,33 @@ type NotifyExt interface {
...
 	Msg          string
 	stack        []uintptr
 	frame        []StackFrame
 	cause        error
 	shouldNotify bool
 	status       *grpcstatus.Status
+	rawData      []interface{}
+}
+
+type Option func(*customError)
```
hothero: I'm mostly thinking if it's good to introduce such good extensibility to this low-level package and does this package needed? error package should be simple and provide basic functionalities to others who want to build up more on top of it, say this pattern is good for the customized struct which built on top of Orion error

what's your thought regarding this?

hothero:
we're also defining a new logging/notifier package and we have key/value arguments for reporting already.

PR #116, opened 2019-08-22
orion error returns grpc status with latest message always

hothero:
the error message got from djgateway calling listing-svc: `cannot create listing on listingsvc :CreateListing :rpc error: code = Internal desc = hystrix: timeout`

you can see the error returned from listing-svc is only `hystrix: timeout` technically but we do wrap every error with custom message in listing-svc. So how it happen without any wrapped message? that's because `Wrap` inherits existing grpc status, so Orion server returns only the message at the beginning as long as you New or Wrap the error as customError.

And now we all use Orion error, so the fix is to enable we always get latest message which is wrapped from customError instead of existing grpc status

### DIFF
--- a/utils/errors/errors.go
+++ b/utils/errors/errors.go
@@ -77,12 +77,15 @@ func (c customError) Cause() error {
 }
 
 func (c customError) GRPCStatus() *grpcstatus.Status {
-	statusCode := codes.Internal
 	if c.status != nil {
-		statusCode = c.status.Code()
+		// use latest error message and keep other data (e.g. details)
+		newStatus := c.status.Proto()
+		newStatus.Message = c.Error()
+
+		return grpcstatus.FromProto(newStatus)
 	}
 
-	return grpcstatus.New(statusCode, c.Error()) // always return with latest error message
+	return grpcstatus.New(codes.Internal, c.Error())
 }
 
 func (c *customError) generateStack(skip int) []StackFrame {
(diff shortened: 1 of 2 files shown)


utils/errors/errors.go
```diff
@@ -77,10 +77,12 @@ func (c customError) Cause() error {
...
 func (c customError) GRPCStatus() *grpcstatus.Status {
-	if c.status == nil {
-		return grpcstatus.New(codes.Internal, c.Error())
+	statusCode := codes.Internal
+	if c.status != nil {
+		statusCode = c.status.Code()
 	}
-	return c.status
+
+	return grpcstatus.New(statusCode, c.Error()) // always return with latest error message
```
benjaminheng: If `c.status` contains any details, the details will be lost when you create a new grpcstatus.Status struct.

hothero: right, I'll update it

utils/errors/errors.go
```diff
@@ -77,10 +77,15 @@ func (c customError) Cause() error {
 }
 
 func (c customError) GRPCStatus() *grpcstatus.Status {
-	if c.status == nil {
-		return grpcstatus.New(codes.Internal, c.Error())
+	if c.status != nil {
+		// use latest error message and keep other data (e.g. details)
+		newStatus := c.status.Proto()
```
hothero: this will clone a new status returned

utils/errors/errors.go
```diff
@@ -77,10 +77,15 @@ func (c customError) Cause() error {
...
+		newStatus.Message = c.Error()
+
+		return grpcstatus.FromProto(newStatus)
```
hothero: return new status with wrapped err msg rather than updating existing status

PR #117, opened 2019-08-26
Migrate Go Modules and remove vendor folder

avence12:
Go Modules have been released for more than one year and teseted for two minor Golang versions.
Let's start to use Modules for other service to integrate Orion easily.

Changes:
- Remove github.com/kardianos/govendor
- Remove `vendor/`
- Downgrade to `go.elastic.co/apm@v1.4.0` for memory leak issue on v1.5.0


.travis.yml
```diff
@@ -2,14 +2,15 @@ language: go
 sudo: false
 
 go:
-  - master
-  - "1.10"
-  - "1.11"
+  - "1.12"
```
itsneo: https://docs.travis-ci.com/user/languages/go/#specifying-a-go-version-to-use
Recommend to use `"1.12.x"` to test against the latest `1.12` version

.gitignore
```diff
@@ -12,3 +12,7 @@
 
 # Project-local glide cache, RE: https://github.com/Masterminds/glide/issues/736
 .glide/
+
+# Go Modules
+go.sum
```
benjaminheng: go.sum should be checked into VCS

https://blog.golang.org/using-go-modules

> The go command uses the go.sum file to ensure that future downloads of these modules retrieve the same bits as the first download, to ensure the modules your project depends on do not change unexpectedly, whether for malicious, accidental, or other reasons. Both go.mod and go.sum should be checked into version control.

.travis.yml
```diff
@@ -2,14 +2,15 @@ language: go
...
 
-install:
-  - go get -u golang.org/x/lint/golint
+env:
+  - GO111MODULE=on
+
+install: true
 
 script:
+  - GO111MODULE=auto go get -u golang.org/x/lint/golint
```
benjaminheng: only the `env` addition is needed. no need to modify the `install` and `script` keys.

https://github.com/carousell/listing/blob/ea4ede5a4e5c1328669508ec252c7fec7fe2b0a5/.travis.yml#L8-L15

avence12: Because Travis build project under $GOPATH and make it as workspace, running the following will make changes to your go.mod.

It may break the dependency revision of your project if you used the same dependency as golint.

```
install:
  - go get -u golang.org/x/lint/golint
```

benjaminheng: makes sense

PR #126, opened 2020-01-14
[CCH-412] Kafka producer

humjosh:
A simple Kafka producer, to make it easy for other services to send backend analytics events.

### DIFF
--- a/utils/kafka/producer_test.go
+++ b/utils/kafka/producer_test.go
@@ -1,0 +1,90 @@
+package kafka
+
+import (
+	"strings"
+	"testing"
+
+	"context"
+
+	"github.com/Shopify/sarama"
+	"github.com/Shopify/sarama/mocks"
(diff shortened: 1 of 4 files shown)


utils/kafka/producer_test.go
```diff
@@ -0,0 +1,90 @@
...
+		asyncProducer.ExpectInputAndSucceed()
+		producer := Producer{
+			asyncProducer: asyncProducer,
+		}
+		producer.Run()
+		defer producer.Close()
+
+		err := producer.Produce(ctx, "test-topic", "test-key", []byte("test-payload"))
+		if err != nil {
+			t.Errorf("expected no error, got %v", err)
```
hrs7991: should replace this with `t.Fatalf()` which ends the test immediately while `t.Errorf()` just marks the test for failure but continues forward (similar comment for other tests)

PR #130, opened 2020-04-01
NewRelic HTTP transaction name

achichen:
# Problem Statement

For HTTP requests, Orion by default tracks it on NewRelic using request url as the transaction name. This works fine for static url such as `/search/products/` while not ideal for dynamic route such as `/search/{username}/products/` which would generate different metric name for each different username.

This PR addresses this by providing a choice for application to decide their HTTP transaction name on NewRelic.

**NOTE: this change only affects HTTP endpoint but not GRPC**

# Proposed Change

Add a new config `newRelicHttpTxName` for application to choose their tx name types:

```
[orion]
newrelicServiceName="your-nr-service-name"
newrelicApiKey="your-nr-api-key"
newRelicHttpTxName="method"
```

Valid `newRelicHttpTxName` values are:
- **method** (default): using the service method name. E.g.  `SearchProducts`, `SearchUserProducts`
- **fullmethod**: using both service and method name. E.g.  `MyService_proto.MyService/SearchProducts`, `MyService_proto.MyService/SearchUserProducts`
- **url**: using the request http method + url. E.g. `POST /search/products/`. And as my example, for different username `POST /search/james/products/` and `POST /search/alex/products/` used.
- **route**: using route . E.g. `POST /search/products/`. And as my example, for different username `POST /search/{username}/products/` is used for all username variation.

Note: On NewRelic UI, transactions name are prefixed with a `/`. Hence **route** tx names, it is displayed as `/POST /search/{username}/products/` on its web UI


orion/config.go
```diff
@@ -73,6 +73,8 @@ type ZipkinConfig struct {
 type NewRelicConfig struct {
 	APIKey            string
 	ServiceName       string
+	//HttpTxName is the transaction name logged in NR. Options are "method" (default), "fullmethod", "url", "route". Default is "method"
+	HttpTxName 	  string
```
cyrandy, line 77: How about `HttpTxNameType`

achichen: Sounds good

interceptors/README.md
```diff
@@ -34,8 +34,11 @@
 * [func HystrixClientInterceptor() grpc.UnaryClientInterceptor](#HystrixClientInterceptor)
```
achichen, line 34: regen by `make gen`

orion/README.md
```diff
@@ -89,7 +89,7 @@ This contains:
 	An interface type for servers to implement, also with the methods defined in the SimpleService (simple.pb.go)
```
achichen, line 89: regen by `make gen`

orion/config.go
```diff
@@ -73,6 +73,8 @@ type ZipkinConfig struct {
 type NewRelicConfig struct {
 	APIKey            string
 	ServiceName       string
+	//HttpTxNameType decides the transaction name logged in NR. Options are "method" (default), "fullmethod", "url", "route". Default is "method"
```
edwardsamuel, line 76: I would suggest to use `fullmethod` as the default since it is same behaviour to [gRPC endpoints](https://github.com/carousell/Orion/blob/07d546eaaf9d99283cb9868b81e440755b1ad3e6/interceptors/interceptors.go#L107). I am also good also to have this as the only option in order to make it consistent across all Orion.
`url` case should not be used, since it makes observability harder.

achichen: Okay to me using `fullmethod` as the default. Will fire another PR to change it.

I feel a valid case (though rare) in using `route` over `method` or `fullmethod` for HTTP requests: Let's say one service exposes the same endpoint in both HTTP and GRPC. Then both traffic will end up being served by the same service method. If we use the same metric name for both, it is not distinguishable on NR.

benjaminheng:
cool, we had the same problem on djgateway, and i ended up using a interceptor to rename the newrelic transaction name.

https://github.com/carousell/djgateway/blob/7a91b4be4bcf3a282fc4873dd38e30100e296a02/djgateway/service/middlewares.go#L116-L129

PR #133, opened 2020-06-18
adding sigint to  handled signals similar to sigterm

vvarma:


orion/core.go
```diff
@@ -311,7 +311,7 @@ func (d *DefaultServerImpl) signalWatcher() {
 				}
 				info.sf.DisposeService(info.ss, params)
 			}
-		} else if sig == syscall.SIGTERM {
+		} else if sig == syscall.SIGTERM || sig == syscall.SIGINT {
```
anaghsys, line 314: Who will be sending SIGINT @vvarma

ankurs: I believe this will be sent on shutdown of the system (also ctrl+c) for graceful termination

PR #134, opened 2020-06-20
add tags support

vizshala:
The purpose is to allow adding tags when reporting warning/error.
This greatly reduces efforts when trying to gather context information.
This change will not affect the existing use of sentry.
For service which wants to make use of sentry tags, just a little modification is required.

- notify with tags
```
	notifier.NotifyWithLevel(ErrorPoslaju(), "warning", "extra string data", notifier.SentryTags{
		"order_id":      "this is order id"
		"user_id":       "this is user id",
		"shipping_code": "this is shipping code",
	})
```

All extra data of type `notifier.SentryTags` will be appended to sentry custom tags.

The effect is demonstrated as follows:
image

### DIFF
--- a/utils/errors/notifier/notifier.go
+++ b/utils/errors/notifier/notifier.go
@@ -22,17 +22,20 @@ import (
 )
 
 var (
-	airbrake      *gobrake.Notifier
-	bugsnagInited bool
-	rollbarInited bool
-	sentryInited  bool
-	serverRoot    string
-	hostname      string
+	airbrake         *gobrake.Notifier
+	bugsnagInited    bool
+	rollbarInited    bool
+	sentryInited     bool
+	sentryTagEnabled bool
+	serverRoot       string
+	hostname         string
 )
 
 const (
 	tracerID = "tracerId"
 )
+
+type Tags map[string]string
 
 // InitAirbrake inits airbrake configuration
 func InitAirbrake(projectID int64, projectKey string) {
@@ -55,6 +58,11 @@ func InitRollbar(token, env string) {
 func InitSentry(dsn string) {
 	raven.SetDSN(dsn)
 	sentryInited = true
+}
+
+func InitSentryWithTagEnabled(dsn string) {
+	InitSentry(dsn)
+	sentryTagEnabled = true
 }
 
 func convToGoBrake(in []errors.StackFrame) []gobrake.StackFrame {
@@ -109,12 +117,18 @@ func convToSentry(in errors.ErrorExt) *raven.Stacktrace {
 	return out
 }
 
-func parseRawData(ctx context.Context, rawData ...interface{}) map[string]interface{} {
+func parseRawData(ctx context.Context, rawData ...interface{}) (map[string]interface{}, map[string]string) {
 	m := make(map[string]interface{})
+	t := make(map[string]string)
 	for pos := range rawData {
 		data := rawData[pos]
 		if _, ok := data.(context.Context); ok {
 			continue
+		}
+		if sentryTags, ok := data.(Tags); ok {
+			for k, v := range sentryTags {
+				t[k] = v
+			}
 		}
 		m[reflect.TypeOf(data).String()+strconv.Itoa(pos)] = data
 	}
@@ -123,7 +137,7 @@ func parseRawData(ctx context.Context, rawData ...interface{}) map[string]interf
 			m[k] = v
 		}
 	}
-	return m
+	return m, t
 }
 
 func Notify(err error, rawData ...interface{}) error {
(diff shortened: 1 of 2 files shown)


utils/errors/notifier/notifier.go
```diff
@@ -109,21 +117,27 @@ func convToSentry(in errors.ErrorExt) *raven.Stacktrace {
...
+func parseRawData(ctx context.Context, rawData ...interface{}) (map[string]interface{}, map[string]string) {
 	m := make(map[string]interface{})
+	t := make(map[string]string)
 	for pos := range rawData {
 		data := rawData[pos]
 		if _, ok := data.(context.Context); ok {
 			continue
 		}
+		if sentryTags, ok := data.(Tags); ok {
+			for k, v := range sentryTags {
```
ankurs, line 129: Will it be possible for us to do this with interfaces ?

vizshala: Do you mean define an interface such as `isTags`
Any struct that implements this interface can be validated and accepted as tags here?

vizshala: Or do you mean assign tags by type assertion without the for loop?

ankurs: First one, something like isTags

vizshala: added isTags for generic tags abstraction.
and a more specific isSentryTags for sentry.

utils/errors/notifier/notifier.go
```diff
@@ -109,21 +128,25 @@ func convToSentry(in errors.ErrorExt) *raven.Stacktrace {
 	return out
 }
 
-func parseRawData(ctx context.Context, rawData ...interface{}) map[string]interface{} {
+func parseRawData(ctx context.Context, rawData ...interface{}) (map[string]interface{}, []interface{}) {
```
benjaminheng, line 131: Use named return parameters here so the method signature is self explanatory. Something like `rawData map[string]interface{}, tags []interface{}`

vizshala:
> What's the reasoning for having two interfaces, `isTags` and `isSentryTags`?
> 
> It's strange that `parseRawData` expects `isTags`, but consumers of `parseRawData` expect `isSentryTags` instead.
> 
> Might as well only have one interface and for `parseRawData` to return `map[string]string` instead. eg.
> 
> interface:
> 
> ```go
> type isTags interface {
>   isTags()
>   GetTags() map[string]string
> }
> ```
> 
> `parseRawData()`:
> 
> ```go
> func parseRawData(ctx context.Context, rawData ...interface{}) (rawData map[string]interface{}, tags map[string]string) {
>     // omitted
>     if v, ok := data.(isTags); ok {
>         tags = v.GetTags()
>     }
> }
> ```
> 
> `Notify()`:
> 
> ```go
> func Notify(err error, rawData ...interface{}) error {
>     // omitted
>     parsedData, tags := parseRawData(ctx, list...)
>     if tags != nil && len(tags) > 0 {
>         packet.AddTags(tags)
>     }
> }
> ```

I agree with you about no need to have more interfaces.
My initial thought is for different tracing services having different tags format.
But anyway for this purpose, the conversion should be done inside this package.
And only a unified tag interface is required.

utils/errors/notifier/notifier.go
```diff
@@ -109,21 +122,25 @@ func convToSentry(in errors.ErrorExt) *raven.Stacktrace {
...
+func parseRawData(ctx context.Context, rawData ...interface{}) (extraData map[string]interface{}, tagData []map[string]string) {
+	extraData = make(map[string]interface{})
+
 	for pos := range rawData {
 		data := rawData[pos]
 		if _, ok := data.(context.Context); ok {
 			continue
 		}
-		m[reflect.TypeOf(data).String()+strconv.Itoa(pos)] = data
+		if tags, ok := data.(isTags); ok {
```
anaghsys, line 133: The tag data should be removed from other info present (in `rawData`).

This will prevent the same info being sent twice - once as part of tag data, once as part of extra data.

vizshala: makes sense.

utils/errors/notifier/notifier.go
```diff
@@ -109,21 +122,26 @@ func convToSentry(in errors.ErrorExt) *raven.Stacktrace {
 	return out
 }
 
-func parseRawData(ctx context.Context, rawData ...interface{}) map[string]interface{} {
-	m := make(map[string]interface{})
+func parseRawData(ctx context.Context, rawData ...interface{}) (extraData map[string]interface{}, tagData []map[string]string) {
```
benjaminheng, line 125: Does `tagData` need to be a slice of maps?

vizshala: for cases when the error wraps another error and each of them has tags respectively.

vizshala: The reason not to unfold and fold all tags into one map takes a for loop is when later `packet.AddTags` was called, it will unfold all tags again.

PR #135, opened 2020-07-02
Fix stream terminated by RST_STREAM with error code: PROTOCOL_ERROR

cyrandy:


orion/handlers/http/http.go
```diff
@@ -20,8 +20,30 @@ import (
...
+		for _, value := range values {
+			if err = handler(key, value); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// Set conforms to the opentracing.TextMapWriter interface.
```
benjaminheng, line 41: add a comment to explain that this carrier ensures metadata keys are lowercased to conform to the HTTP/2 spec

benjaminheng:
Adding some context to the fix.

Previously when the following is used:

```
opentracing.GlobalTracer().Inject(wireContext, opentracing.HTTPHeaders, opentracing.HTTPHeadersCarrier(md))
```

The gRPC metadata is wrapped with a `opentracing.HTTPHeadersCarrier`. Under the hood this carrier treats itself as a `http.Header` struct before calling `.Add()` on itself:

```
func (c HTTPHeadersCarrier) Set(key, val string) {
	h := http.Header(c)
	h.Add(key, val)
}
```

`.Add()` on the `http.Header` type canonicalizes the key name to with uppercased words. Eg. `accept-language` -> `Accept-Language`. This is not allowed by the HTTP/2 spec (used by gRPC), and results in the following error when making a gRPC call to a downstream service:

```
Fix stream terminated by RST_STREAM with error code: PROTOCOL_ERROR
```

More specifically, the zipkin tracer is injecting headers like `x-b3-sampled: 0`, which is getting changed to `X-B3-Sampled: 0`.

orion/handlers/http/http.go
cyrandy: added

PR #137, opened 2020-08-05
Config to disable default interceptors

achichen:
This PR is allow service to disable Orion's [default interceptors](https://github.com/carousell/Orion/blob/master/interceptors/interceptors.go#L39-L49) by setting config

```
[orion]
NoDefaultInterceptors=true
```

This could be useful to disable interceptors such as `ResponseTimeLoggingInterceptor` which logs massively for each endpoint calls.


orion/config.go
```diff
@@ -51,6 +51,8 @@ type Config struct {
 	Env string
 	// DefaultJSONPB sets jsonpb as the encoder/decoder for application/json request/response bodies
 	DefaultJSONPB bool
+	// NoDefaultInterceptors disables all default interceptors on all handlers
+	NoDefaultInterceptors bool
```
anaghsys, line 55: Can we rename this to `EnableDefaultInterceptors` with a default value of `true`? Similarly, can we change the config to be `orion.EnableDefaultInterceptors` to better convey the type of the config.

benjaminheng: I don't agree with this. I think the empty value (false for a boolean) should be the default, and the naming of the flag should accommodate the default value.

e.g. if by default we want interceptors, the flag should be named `DisableInterceptors`, on the flipside if we wanted no interceptors by default, then the flag should be named `EnableInterceptors`.

benjaminheng:
Overall I'm ok with the change, since Orion already supported the flag, it just wasn't exposed.

However there are some interceptors that are quite core to our ecosystem (opentracing, prometheus, error reporting). Allowing services to disable *all* interceptors is putting the responsibility on the service owners to properly re-specify these core interceptors. If not done properly, their service wouldn't get metrics, or wouldn't be traced, or wouldn't get errors reported to sentry, etc.

orion/config.go
anaghsys: You're correct @benjaminheng . My suggestion was about renaming the var - the final name of course is up to the logic you've mentioned.

So in this case, `DisableDefaultInterceptors` with a `false` value would be good.

anaghsys:
Another approach that this PR can take is one of filtered interceptors: give names to the interceptors currently in use and then allow services to filter them as needed. This will allow for flexibility and not lead to loss of core interceptors if the service doesn't actively remove them.

This will take the approach out of config and make it code-driven, but (IMO) rarely do you need to change interceptors in the blink of a moment via config files.

orion/config.go
achichen: For the config name, I'm also in favor of `DisableDefaultInterceptors`

achichen:
I agree with @benjaminheng 's point on this is putting responsibility to service owners. But I feel that what Orion should do.

Orion is aiming to be an open source framework. Though framework can be opinionated, ideally the default setting should be as general as possible.

I am not exactly sure whether `ResponseTimeLoggingInterceptor`, `filterFromZipkin`, `grpc_prometheus.UnaryServerInterceptor`, as well as `NewRelicInterceptor` as the default are too Carousell-specific. That means anyone outside Carousell using this framework must operate on the same environment as us. And even our own environment might change, for example, we even decided to reduce NewRelic usage not long ago. To enforce it as a framework default seems to less flexible.

IMO, it's good for Orion to support these popular integrations out of box. But it should be the skeleton tool (ie. Orion-Builder) responsibility to standardize what to be use within Carousell. Orion itself should only default a set of interceptors, which are standard (e.g. opentracing) or common pattern (e.g. PanicHandler). And it's the user's preference to extend it by setting up their own `GetInterceptors()`.

benjaminheng:
Hmm alright, sounds reasonable to me. I'm good with this change then

hothero:
overall I'm good with this change as we've had that config but just it was not exposed yet. However, in the long-run we still need more flexible configuration and documentation to let engineers understand what's the downside of this config and what's the responsibility that they need to take with disabling the DefaultServerInterceptor

orion/config.go
hothero: @achichen what do you think if we also separate the config for unary and stream interceptors? https://github.com/carousell/Orion/blob/master/orion/handlers/utils.go#L135 as they also have different interface functions to fill in back. In general, they may have different needs.

achichen: @benjaminheng @anaghsys I've renamed the config to `DisableDefaultInterceptors`

praveensanap:
@achichen @benjaminheng @hothero @anaghsys 
Seems that we have consensus on the PR. 
If no more comments, can you guys approve the PR?

hothero (approved):

PR #138, opened 2020-08-06
Omit healthcheck traces

benjaminheng:
Some services define their healthcheck RPC with alternate casing.


benjaminheng:
Build is failing because https://gitlab.howett.net/go/plist.git/ is unreachable at the moment. Will build again later 🙄

PR #139, opened 2020-08-31
Add http code to notify in http handler

kylelwm:
Add "http_code" tag when http handler notifies error

This is to allow filtering of errors by status code (400, 401, 500) on sentry

image

### DIFF
--- a/orion/handlers/http/http.go
+++ b/orion/handlers/http/http.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


orion/handlers/http/http.go
```diff
@@ -65,7 +66,17 @@ func (h *httpHandler) httpHandler(resp http.ResponseWriter, req *http.Request, s
...
 		utils.FinishNRTransaction(req.Context(), nil)
 	} else {
-		notifier.Notify(err, req.URL.String(), ctx)
+
+		// Add HTTP response code as tag
+		var tags notifier.Tags
+		if err != nil {
+			httpCode, _ := GrpcErrorToHTTP(err, http.StatusInternalServerError, "Internal Server Error!")
+			tags = notifier.Tags{
+				"http_code": strconv.FormatInt(int64(httpCode), 10),
```
benjaminheng, line 75: `httpCode` is an int, so you can use `strconv.Itoa` instead.

benjaminheng (approved):

orion/handlers/http/http.go
kylelwm: :salute:

PR #142, opened 2020-09-18
Add grpc_retry.UnaryClientInterceptor() as a default client interceptor

humjosh:
Adding `grpc_retry.UnaryClientInterceptor()` to `DefaultClientInterceptors` makes it easy for services to enable retries for gRPC method calls. By default, the number of retries is zero, so this change has no effect on services unless they opt in to the behaviour.

Services will most commonly use these options to configure the retry behaviour:
* `grpc_retry.WithMax` - specifies the number of retries
* `grpc_retry.WithCodes` - specifies which gRPC error codes to retry on (defaults to `ResourceExhausted` and `Unavailable`)

Simplest usage example:
```go
import (
	...
	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
)

func (c *client) GetThing (ctx context.Context, req *GetThingRequest) (*GetThingResponse, error) {
	return c.thingClient.GetThing(ctx, req, grpc_retry.WithMax(3))
}
```

I have added it as the _first_ interceptor, so that it wraps around the tracing interceptor.

### DIFF
--- a/builder/vendor/github.com/carousell/Orion/interceptors/interceptors.go
+++ b/builder/vendor/github.com/carousell/Orion/interceptors/interceptors.go
@@ -13,9 +13,8 @@ import (
 	"github.com/carousell/Orion/utils/errors/notifier"
 	"github.com/carousell/Orion/utils/log"
 	"github.com/carousell/Orion/utils/log/loggers"
-	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
-	grpc_retry "github.com/grpc-ecosystem/go-grpc-middleware/retry"
-	grpc_ctxtags "github.com/grpc-ecosystem/go-grpc-middleware/tags"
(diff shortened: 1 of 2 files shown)


ankurs:
This looks like a good addition, my only concern is how will the teams be able to specify specific params like Backoff or codes to retry on ?

humjosh:
> This looks like a good addition, my only concern is how will the teams be able to specify specific params like Backoff or codes to retry on ?

Howdy @ankurs !

The `grpc_retry` package offers options for configuring backoff and codes to retry on. See here: https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/retry#CallOption

benjaminheng (changes requested):
The vendored code in the `builder/` directory is being modified, instead of `interceptors/interceptors.go`

benjaminheng (approved):

PR #143, opened 2020-10-05
Enforce timeout on GracefulStop in grpcHandler

harshadr:
Active stream connections would block GracefulStop preventing process shutdown upon SIGINT.


benjaminheng (approved):

edwardsamuel (approved):
Looks good to me. 
`g.grpcServer.Stop()` can interrupt `g.grpcServer.GracefulStop()`.

PR #146, opened 2020-12-22
modify error tag to fit opentracing convention

stevenCarousell:
According to the opentracing tags spec, the error should be boolean.
https://github.com/opentracing/specification/blob/master/semantic_conventions.md#span-tags-table

The vendor e.g Jaeger will treat the tag as a special render on jaeger-query page.
It will be easier to find error spans by filter the error=true tag also.

Screen Shot 2020-12-22 at 1 50 15 PM
Screen Shot 2020-12-22 at 9 59 40 PM

### DIFF
--- a/utils/spanutils/spanutils.go
+++ b/utils/spanutils/spanutils.go
@@ -83,7 +83,8 @@ func (span *tracingSpan) SetError(msg string) {
 		return
 	}
 	if msg != "" {
-		span.openSpan.SetTag("error", msg)
+		span.openSpan.SetTag("error", true)
+		span.openSpan.SetTag("error.msg", msg)
 	}
 }
 


benjaminheng (commented):
You're changing the vendored code by mistake

benjaminheng:
Also, set it to `true` not `"true"`.

stevenCarousell:
> You're changing the vendored code by mistake

thanks for point out, just see it's in vendor.

utils/spanutils/spanutils.go
```diff
@@ -83,7 +83,8 @@ func (span *tracingSpan) SetError(msg string) {
 		return
 	}
 	if msg != "" {
-		span.openSpan.SetTag("error", msg)
+		span.openSpan.SetTag("error", true)
+		span.openSpan.SetTag("error.msg", msg)
```
hothero, line 87: according to [Jaeger document](https://opentracing.io/docs/overview/tags-logs-baggage/), they don't mention anything about the size difference between Tag and Log. I think based on our needs, Tag would be better than using Log. It looks good for me.

hothero: I will create another PR to use SetError in every endpoint call.

hothero (approved):

PR #147, opened 2021-01-06
Create notifier level types and do logging with correct level

hothero:
notifier allows users to do NotifyWithLevel and the argument is string type, that means the possibility to have human error. It would be great to define our own level type for people who want to use notifier package. 

Also, in this PR, we refactor a little bit the doNotify function to have severity defined at the beginning, so subsequent logic can use it directly and prevent abusing

### DIFF
--- a/utils/errors/notifier/notifier.go
+++ b/utils/errors/notifier/notifier.go
@@ -171,6 +171,7 @@ func doNotify(err error, skip int, level string, rawData ...interface{}) error {
 	if err == nil {
 		return nil
 	}
+	sev := parseLevel(level)
 
 	// add stack infomation
 	errWithStack, ok := err.(errors.ErrorExt)
@@ -242,16 +243,10 @@ func doNotify(err error, skip int, level string, rawData ...interface{}) error {
 			fields = append(fields, &rollbar.Field{Name: "traceId", Data: traceID})
 		}
 		fields = append(fields, &rollbar.Field{Name: "server", Data: map[string]interface{}{"hostname": getHostname(), "root": getServerRoot()}})
-		rollbar.ErrorWithStack(level, errWithStack, convToRollbar(errWithStack.StackFrame()), fields...)
+		rollbar.ErrorWithStack(sev.String(), errWithStack, convToRollbar(errWithStack.StackFrame()), fields...)
 	}
 
 	if sentryInited {
-		defLevel := raven.ERROR
-		if level == "critical" {
-			defLevel = raven.FATAL
-		} else if level == "warning" {
-			defLevel = raven.WARNING
-		}
 		ravenExp := raven.NewException(errWithStack, convToSentry(errWithStack))
 		packet := raven.NewPacketWithExtra(errWithStack.Error(), parsedData, ravenExp)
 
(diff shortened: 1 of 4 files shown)


utils/errors/notifier/notifier.go
```diff
@@ -242,28 +243,22 @@ func doNotify(err error, skip int, level string, rawData ...interface{}) error {
...
 			packet.AddTags(tags)
 		}
 
-		packet.Level = defLevel
+		packet.Level = sev.RavenSeverity()
 		raven.Capture(packet, nil)
 	}
 
-	log.GetLogger().Log(ctx, loggers.ErrorLevel, skip+1, "err", errWithStack, "stack", errWithStack.StackFrame())
+	log.GetLogger().Log(ctx, sev.LoggerLevel(), skip+1, "err", errWithStack, "stack", errWithStack.StackFrame())
```
hothero, line 261: now we can do logging with correct level

utils/errors/notifier/types.go
```diff
@@ -0,0 +1,34 @@
...
+
+const (
+	ErrorLevel   = "error"
+	DebugLevel   = "debug"
+	InfoLevel    = "info"
+	WarningLevel = "warning"
+	FatalLevel   = "fatal"
+)
+
+const (
```
hothero, line 16: inspired from https://github.com/getsentry/raven-go/blob/master/client.go#L45-L54

utils/errors/notifier/utils.go
```diff
@@ -0,0 +1,36 @@
...
+	switch s {
+	case warningSeverity:
+		return loggers.WarnLevel
+	case infoSeverity:
+		return loggers.InfoLevel
+	case debugSeverity:
+		return loggers.DebugLevel
+	default:
+		return loggers.ErrorLevel
+	}
```
benjaminheng, line 26: Add

```
case errorSeverity:
  return loggers.ErrorLevel
```

instead of relying on the default case to handle the error severity.

So that if someone changes the default case in the future for some reason, they won't accidentally affect calls that are using `errorSeverity`

hothero: fair point, updated.

utils/errors/notifier/notifier.go
```diff
@@ -242,28 +243,22 @@ func doNotify(err error, skip int, level string, rawData ...interface{}) error {
...
-		}
 		ravenExp := raven.NewException(errWithStack, convToSentry(errWithStack))
 		packet := raven.NewPacketWithExtra(errWithStack.Error(), parsedData, ravenExp)
 
 		for _, tags := range tagData {
 			packet.AddTags(tags)
 		}
 
-		packet.Level = defLevel
+		packet.Level = sev.RavenSeverity()
```
edwardsamuel, line 257: Instead of doing `sev.RavenSeverity()`, IMO, it would be better to expose a func that convert `sev` object into `raven.Severity` in this package. I don't think the `.RavenSeverity` is in use other than here (ie: when we want to send it to Sentry)

hothero: maintaining the wrapper for converting types as util func instead of as a struct function sounds good to me.

hothero: In the end, I just removed the struct function and do type conversion directly since it's just a single use case. 🙏

benjaminheng (approved):

PR #149, opened 2021-02-18
add ForwardMetadata as one of default interceptors

hothero:
Metadata is one of the important concepts while communicating multiple microservices, so it's good to have it as default interceptor when calling to other services.

### DIFF
--- a/interceptors/interceptors.go
+++ b/interceptors/interceptors.go
@@ -3,6 +3,7 @@ package interceptors
 import (
 	"context"
 	"fmt"
+	"google.golang.org/grpc/metadata"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


interceptors/interceptors.go
```diff
@@ -209,6 +211,23 @@ func HystrixClientInterceptor() grpc.UnaryClientInterceptor {
 	}
 }
 
+// ForwardMetadataInterceptor forwards metadata from upstream to downstream
+func ForwardMetadataInterceptor() grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		md, ok := metadata.FromIncomingContext(ctx)
+		if ok {
+			// means that we have some incoming context values needed to pass through following services
+			// e.g. fulfillgw -> caroorders -> caroship -> caroseveneleven
```
benjaminheng, line 220: replace service names with generic ones

benjaminheng (approved):

interceptors/interceptors.go
hothero: right, good point. Thx.

humjosh (approved):

PR #150, opened 2021-03-03
reorder defaultclientinterceptors

hothero:
A request that failed could be many reasons including network-level issues, connection re-establishment, etc.. Hystrix as implementing circuit breaker pattern considers how to avoid intermittent network hiccups as well. Also, we don't want to have circuit open because of too many retries that were probably intended.

In other words, we consider Hystrix Command as a whole for a request to other services regardless of whether retry or not.

However, it may be a breaking change to some services that have been using grpc_retry options. We will have to notice those services after we merge this PR.

### DIFF
--- a/interceptors/interceptors.go
+++ b/interceptors/interceptors.go
@@ -53,10 +53,10 @@ func DefaultInterceptors() []grpc.UnaryServerInterceptor {
 //DefaultClientInterceptors are the set of default interceptors that should be applied to all client calls
 func DefaultClientInterceptors(address string) []grpc.UnaryClientInterceptor {
 	return []grpc.UnaryClientInterceptor{
+		HystrixClientInterceptor(),
 		grpc_retry.UnaryClientInterceptor(),
 		GRPCClientInterceptor(),
 		NewRelicClientInterceptor(address),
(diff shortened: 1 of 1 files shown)


benjaminheng:
I don't agree with this. The reason for a circuit breaker is to start failing requests based on the how many outgoing requests are failing.

> Also, we don't want to have circuit open because of too many retries that were probably intended.

We do want that - each retry attempt is a request that the downstream service has to handle. If a service is saturating another service with many failing requests, we should be opening the circuit breaker regardless of whether those requests are retries or not. Allowing services to retry indefinitely within a single hystrix command kind of defeats the purpose of a circuit breaker.

Basically: **I think retries should be treated the same as the first request. i.e. they should contribute just as much to the circuit breaker's health as the first request.**

ankurs:
I agree with @benjaminheng that was the original intention

hothero:
@ankurs @benjaminheng thanks for your comments. I was confused about some reading regarding the two patterns. After re-read them and your comments, I think the opinion is reasonable. If we changed the order, it could generate some more unexpected retries as well. I'm going to close this.

PR #152, opened 2021-04-03
init incoming context as default for HTTP server

hothero:
We've onboarded forward metadata via incoming and outgoing context in https://github.com/carousell/Orion/pull/149, but in HTTP server we haven't inited an incoming context inside, which means we can use the metadata in the incoming context that will be transformed to outgoing when we call to other services.

Also, in our current interface of encoders: https://github.com/carousell/Orion/blob/master/orion/handlers/types.go#L40-L44, it doesn't support context assigning back, which will cause new context with incoming context lost. We will have to init an incoming context in preparation stage. (Btw, for gRPC server, the context always has incoming context regardless)

### DIFF
--- a/orion/handlers/http/http.go
+++ b/orion/handlers/http/http.go
@@ -99,6 +99,9 @@ func prepareContext(req *http.Request, info *methodInfo) context.Context {
 	//initialize headers
 	ctx = headers.AddToRequestHeaders(ctx, "", "")
 	ctx = headers.AddToResponseHeaders(ctx, "", "")
+	// initialize incoming context as gRPC server has it naturally
+	// so we can have features where we utilize it for both types of server
+	ctx = metadata.NewIncomingContext(ctx, metadata.MD{})
 
 	// fetch and populate whitelisted headers
 	if len(info.svc.requestHeaders) > 0 {


hothero:
@rajathishere @edwardsamuel @achichen @benjaminheng what do you think about this change? In the long-run, I think we will need to refactor the Orion structure to make gRPC as a base and HTTP is just a layer where we have additional HTTP protocol features upon it. This PR is related to the vision but just a small step to the goal.

achichen:
> @rajathishere @edwardsamuel @achichen @benjaminheng what do you think about this change? In the long-run, I think we will need to refactor the Orion structure to make gRPC as a base and HTTP is just a layer where we have additional HTTP protocol features upon it. This PR is related to the vision but just a small step to the goal.

@hothero I would like to clarify the need. 

Is this a necessary change? Because today without this initialization, we still [create a new incoming context in `AppendHTTPHeadersToIncomingContextInterceptor` from cheader package](https://github.com/carousell/go-utils/blob/master/cheaders/headers.go#L23). This approach seems nicer to me, where only when grpc incoming context is initialized only when it is needed.

My concern are
- It is unnecessarily tightly couples http to grpc (I admit that we have done so by building http over gRPC in Orion. so it's not really a serious concern)
- To a user who uses Orion to build http service, Orion "magically" creates an empty gRPC incoming context on the arrival of a http request, but it is not propagated anything and doesn't bring any values.

So I would like to understand more your vision on this: what's the possible usage/benefit if we initialize grpc context from the very beginning?

hothero:
@achichen I understand the concern, my thought was mostly coming from a structure point of view. So, let's say the structure would look like in the long-run:
```
                     HTTP requests.                   gRPC requests
                            (   service functions (with context)   )
```
so engineers as the owner of service functions will know no matter whether the requests are from HTTP or gRPC they have the same context values to operate (including incoming context). 

I thought that would be nice to all have "incoming context" initialized. However, we can make it more meaningful. Saying that  HTTP's headers are the gRPC's metadata. We could just put all HTTP headers into incoming context to have the same idea. Currenly, in Orion we forward every incoming as outgoing metadata, and that makes me concerning a few sensitive headers such as Authentication token. I think even though we want to do the structure mentioned above, we also need to refactor DefaultClientInterceptor. It lacks a comprehensive plan now. I will close this PR for now.


> Is this a necessary change? Because today without this initialization, we still [create a new incoming context in AppendHTTPHeadersToIncomingContextInterceptor from cheader package](https://github.com/carousell/go-utils/blob/master/cheaders/headers.go#L23). This approach seems nicer to me, where grpc incoming context is only initialized only when it is needed.

However, as maintaining an open-source project, we may need to think about a more generic approach to the Orion project. Then see if anything impacts our internal projects, and we just need to tackle our internal challenges.

achichen:
> so engineers as the owner of service functions will know no matter whether the requests are from HTTP or gRPC they have the same context values to operate (including incoming context).

If we agree Orion should continue exposing unified development experience for gRPC and HTTP, this change makes sense. But I personally feel there are inherently differences between the two protocols hence they will always be different. Say, HTTP has query string and path parameter, which have no mapping concept in gRPC. Should they be part of the "incoming context" of a HTTP request? If so, should we also use grpc metadata context as the holder?

> Currenly, in Orion we forward every incoming as outgoing metadata, and that makes me concerning a few sensitive headers such as Authentication token. I think even though we want to do the structure mentioned above, we also need to refactor DefaultClientInterceptor

I feel `ForwardMetadataInterceptor` is a useful tool. Automatically forwarding incoming metadata to outgoing calls shouldn't be a need specific to Carousell. We can opinionatedly put in as default client interceptors, and other users can choose not to use our default if they don't like it.

To propagate http headers (and exclude certain ones) into incoming context, in Carousell, we solve it using `AppendHTTPHeadersToIncomingContextInterceptor`. We can generalize it and include it into with Orion's http module and user can opt it in.

But anyway, as you've closed this PR. We can discuss this separately.  

cc @benjaminheng you might be interested in this topic regarding Orion http support

PR #154, opened 2021-05-20
add headers only when it presents in request

hothero:
Add headers into context when it presents in request only

Before this change, if we have the allow list ["platform", "authentication", "build-no"]
even though you didn't pass anyone of them, it still has empty string in context.

After this change, we will only have headers in context only when it presents in request. Which makes more sense since we do business logic only when requests present in request.

### DIFF
--- a/orion/handlers/http/http.go
+++ b/orion/handlers/http/http.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/textproto"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


orion/handlers/http/http.go
```diff
@@ -103,7 +104,13 @@ func prepareContext(req *http.Request, info *methodInfo) context.Context {
 	// fetch and populate whitelisted headers
 	if len(info.svc.requestHeaders) > 0 {
 		for _, hdr := range info.svc.requestHeaders {
-			ctx = headers.AddToRequestHeaders(ctx, hdr, req.Header.Get(hdr))
+			if values, found := req.Header[textproto.CanonicalMIMEHeaderKey(hdr)]; found {
```
hothero, line 107: req.Header.Get involves CanonicalMIMEHeaderKey so we use the same format accessing headers map

benjaminheng (approved):

PR #155, opened 2021-05-29
support hystrix error suppressor

hothero:
While introducing DefaultClientInterceptor in more scenarios, there is a case we need an extension in HystrixInterceptor to suppress some specific errors to not count as hystrix circuit. 

For example, when we use HystrixInterceptor for auth-svc or authentication-related logics, with retry mechanism we may not want to count UnAuthenticate, NotFound as Error in hystrix. Because of high QPS, it could cause hystrix circuit-break easier.

### DIFF
--- a/interceptors/interceptors.go
+++ b/interceptors/interceptors.go
@@ -188,9 +188,6 @@ func HystrixClientInterceptor() grpc.UnaryClientInterceptor {
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 		options := clientOptions{
 			hystrixName: method,
-			hystrixErrorSuppressor: func(e error) error {
-				return e
-			},
 		}
 		for _, opt := range opts {
 			if opt != nil {
@@ -209,11 +206,15 @@ func HystrixClientInterceptor() grpc.UnaryClientInterceptor {
 					log.Error(ctx, "panic", r, "method", method, "req", req, "reply", reply)
 				}
 			}()
-			defer notifier.NotifyOnPanic(newCtx, method)
 			// error assigns back to the err object out of hystrix anyway
 			err = invoker(newCtx, method, req, reply, cc, opts...)
-			// error could be suppressed in hystrix wrapper by the function from options
-			return options.hystrixErrorSuppressor(err)
+			defer notifier.NotifyOnPanic(newCtx, method)
+			if options.hystrixErrorSuppressor != nil {
+				// error could be suppressed in hystrix wrapper by the function from options
+				return options.hystrixErrorSuppressor(err)
+			} else {
+				return err
+			}
 		}, nil)
 
 		return err
--- a/interceptors/interceptors.go
+++ b/interceptors/interceptors.go
@@ -3,9 +3,10 @@ package interceptors
 import (
 	"context"
 	"fmt"
-	"google.golang.org/grpc/metadata"
 	"strings"
 	"time"
+
+	"google.golang.org/grpc/metadata"
 
 	"github.com/afex/hystrix-go/hystrix"
 	"github.com/carousell/Orion/orion/modifiers"
@@ -186,12 +187,12 @@ func GRPCClientInterceptor() grpc.UnaryClientInterceptor {
 //HystrixClientInterceptor is the interceptor that intercepts all cleint requests and adds hystrix info to them
 func HystrixClientInterceptor() grpc.UnaryClientInterceptor {
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
-		options := clientOptions{
-			hystrixName: method,
+		options := hystrixOptions{
+			cmdName: method,
 		}
 		for _, opt := range opts {
 			if opt != nil {
-				if o, ok := opt.(clientOption); ok {
+				if o, ok := opt.(hystrixOption); ok {
 					o.process(&options)
 				}
 			}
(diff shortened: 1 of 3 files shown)


interceptors/interceptors.go
```diff
@@ -198,16 +201,22 @@ func HystrixClientInterceptor() grpc.UnaryClientInterceptor {
...
 					err = errors.Wrap(fmt.Errorf("Panic inside hystrix Method: %s, req: %v, reply: %v", method, req, reply), "Hystrix")
 					log.Error(ctx, "panic", r, "method", method, "req", req, "reply", reply)
 				}
 			}()
 			defer notifier.NotifyOnPanic(newCtx, method)
-			return invoker(newCtx, method, req, reply, cc, opts...)
+			// error assigns back to the err object out of hystrix anyway
+			err = invoker(newCtx, method, req, reply, cc, opts...)
+			// error could be suppressed in hystrix wrapper by the function from options
+			return options.hystrixErrorSuppressor(err)
```
hothero, line 216: we could have a function like
```
func (er error) error {
    if e, ok := status.FromError(er); ok {
	    switch e.Code() {
	    // Don't register hystrix errors for the following errors
	    case codes.Canceled, codes.Unauthenticated, codes.NotFound:
                  return nil
	    }
    }
}
```

interceptors/interceptors.go
```diff
@@ -188,6 +188,9 @@ func HystrixClientInterceptor() grpc.UnaryClientInterceptor {
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 		options := clientOptions{
 			hystrixName: method,
+			hystrixErrorSuppressor: func(e error) error {
+				return e
+			},
```
benjaminheng, line 193: leave it as nil

interceptors/interceptors.go
benjaminheng, line 216: Instead of treating `hystrixErrorSuppressor` as always present and relying on the function to be implicitly initialized as a passthrough, let's only call `hystrixErrorSuppressor` if non-nil. Basically if `hystrixErrorSuppressor` exists, then it was provided by the caller.

```
if options.hystrixErrorSupressor != nil {
  return options.hystrixErrorSupressor(invoker(newCtx, method, req, reply, cc, opts...))
} else {
  return invoker(newCtx, method, req, reply, cc, opts...)
}
```

interceptors/options.go
```diff
@@ -30,3 +31,14 @@ func WithHystrixName(name string) clientOption {
 		},
 	}
 }
+
+// WithHystrixErrorSuppressor applies a function that you can write logics to suppress errors to hystrix interceptor
+func WithHystrixErrorSuppressor(e func(err error) error) clientOption {
```
edwardsamuel, line 36: Actually, we have some Hystrix options in https://github.com/carousell/go-utils/blob/master/utils/hystrixwrapper/hystrix_opts.go. I think it might be good if we can utilise this function here.

edwardsamuel: https://github.com/carousell/go-utils is a private repo. Can't be use here.

interceptors/interceptors.go
hothero: yea, I was doing this way and I thought defining a default Suppressor could make the code cleaner. However, I think checking non-nil is better for readability. I will make the change.

hothero: changed and added one unit test https://github.com/carousell/Orion/pull/155/commits/282124e021c025599fa01e4c519a76fd460a779e#diff-c52dbca06ba660883cf2e48f7e11d68dcab4d17b453d068837aff683f3de61cbR1

interceptors/options.go
```diff
@@ -9,6 +9,7 @@ type clientOption interface {
 
 type clientOptions struct {
 	hystrixName string
+	hystrixErrorSuppressor func(err error)error
```
edwardsamuel, line 12: There should be a space in between `func(err error) error`

interceptors/interceptors.go
```diff
@@ -198,16 +198,26 @@ func HystrixClientInterceptor() grpc.UnaryClientInterceptor {
 		}
 		newCtx, cancel := context.WithCancel(ctx)
 		defer cancel()
-		return hystrix.Do(options.hystrixName, func() (err error) {
+		var err error
+		_ = hystrix.Do(options.hystrixName, func() (e error) {
```
edwardsamuel, line 202: No need to name the error as `e` if not in used

interceptors/options.go
```diff
@@ -30,3 +31,14 @@ func WithHystrixName(name string) clientOption {
 		},
 	}
 }
+
+// WithHystrixErrorSuppressor applies a function that you can write logics to suppress errors to hystrix interceptor
```
edwardsamuel, line 35: Would be good if you can give a sample of this

benjaminheng (commented):
do you think it'll be better to keep a similar signature to our internal go-utils version?

```
func WithIgnorableGrpcCode(a ...codes.Code) HystrixOption
```

as opposed to needing callers to define a cumbersome function like:

```
func (er error) error {
    if e, ok := status.FromError(er); ok {
	    switch e.Code() {
	    // Don't register hystrix errors for the following errors
	    case codes.Canceled, codes.Unauthenticated, codes.NotFound:
                  return nil
	    }
    }
}
```

hothero:
@edwardsamuel @benjaminheng plz help review it again, I refactored the whole structure to support more hystrix options.

interceptors/options.go
```diff
@@ -1,32 +1,83 @@
...
-type clientOption interface {
+type hystrixOption interface {
 	grpc.CallOption
-	process(*clientOptions)
+	process(*hystrixOptions)
 }
 
-type clientOptions struct {
-	hystrixName string
+type hystrixOptions struct {
```
benjaminheng, line 15: i think `clientOptions` is probably more appropriate since the options may not necessarily be hystrix related in the future, but not a blocker since it's a private implementation detail

benjaminheng (approved):

interceptors/options.go
hothero: I thought about it, but I would prefer changing it to hystrix-specific for now because I can't think of any use case for other existing interceptors to utilize those options yet. So in the future, if we need generic options for most interceptors, the clientOptions would be a good timing to show up again then.

PR #158, opened 2021-07-26
Add new/wrap error funcs with gRPC code provided

changchingchen:
Add new/wrap error function where user can provide with gRPC code.

### DIFF
--- a/utils/errors/errors.go
+++ b/utils/errors/errors.go
@@ -157,6 +157,12 @@ func NewWithSkipAndStatus(msg string, skip int, status *grpcstatus.Status) Error
 	return WrapWithSkipAndStatus(fmt.Errorf(msg), "", skip+1, status)
 }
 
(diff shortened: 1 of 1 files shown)


utils/errors/errors.go
```diff
@@ -172,6 +178,15 @@ func WrapWithSkip(err error, msg string, skip int) ErrorExt {
 	return WrapWithSkipAndStatus(err, msg, skip+1, nil)
 }
 
+// WrapWithGRPCCode wraps an existing error and appends stack information if it does not exists along with GRPC status with the given GRPC code
```
hothero, line 181: it looks ok to me. I was thinking about putting this more detailed use case in every service and maybe an organization can create an internal package to wrap it up. But as Carousell is an example, we usually directly use Orion's error package. I think in this small case, it's nice to let Orion has the convenience. @edwardsamuel any concern?

hothero (approved):

PR #163, opened 2021-07-29
Forward headers package

hothero:
The main idea is to have the capability to whitelist outgoing headers and not break our DefaultClientInterceptor.
* Create a forwardheaders package responsible for server and client interceptors, allowlist management through context and some util functions
* Extend orion.config to have forwardheaders as the allowlist
* Extend orion.config to have NoDefaultInterceptors (we have the capability but seems no one set it yet)
* Extend DefaultServerInterceptors to set allowlist for forwardheaders
* Transform HTTP headers into gRPC headers (metadata)

### DIFF
--- a/utils/forwardheaders/interceptors.go
+++ b/utils/forwardheaders/interceptors.go
@@ -41,7 +41,7 @@ func UnaryClientInterceptor() grpc.UnaryClientInterceptor {
 	}
 }
 
-// ForwardMetadataInterceptor forwards metadata from upstream to downstream
+// StreamClientInterceptor forwards metadata from upstream to downstream
 func StreamClientInterceptor() grpc.StreamClientInterceptor {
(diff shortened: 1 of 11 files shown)


utils/forwardheaders/interceptors.go
```diff
@@ -0,0 +1,83 @@
...
+	"google.golang.org/grpc/metadata"
+)
+
+func forwardHeadersThroughCtx(ctx context.Context) context.Context {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if ok {
+		// means that we have some incoming context values needed to pass through following services
+		// e.g. api-gateway -> service1 -> service2
+		allowlist := AllowlistFromContext(ctx)
+		if allowlist == nil {
```
hothero, line 16: mainly for backward compatability

hothero:
as leaning to let clients own this responsibility, closing this PR for now.

PR #165, opened 2021-09-01
correctly propagate the hystrix error

stevenCarousell:
Propagate the hystrix error e.g timeout / circuit open from interceptor properly.

https://github.com/afex/hystrix-go/blob/master/hystrix/hystrix.go#L250-L252

### DIFF
--- a/interceptors/unary_interceptors.go
+++ b/interceptors/unary_interceptors.go
@@ -158,7 +158,7 @@ func HystrixClientInterceptor() grpc.UnaryClientInterceptor {
 			}
 		}, options.fallbackFunc)
 		if herr != nil {
-			err = herr
+			return herr
 		}
 		return err
 	}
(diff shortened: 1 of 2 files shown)


interceptors/unary_interceptors.go
```diff
@@ -141,7 +141,7 @@ func HystrixClientInterceptor() grpc.UnaryClientInterceptor {
 		newCtx, cancel := context.WithCancel(ctx)
 		defer cancel()
 		var err error
-		_ = hystrix.Do(options.cmdName, func() (e error) {
+		herr := hystrix.Do(options.cmdName, func() (e error) {
```
hothero, line 144: could you also add unit tests to ensure we can get errors from hystrix itself https://github.com/carousell/Orion/blob/92b31a1e1157fd7355ee449a91c9307cb464392e/interceptors/interceptors_test.go#L36

hothero (approved):

PR #170, opened 2021-12-09
Addition of custom http grpc handlers to support reverse proxy by Orion

vikrammais:


NirbhayK (approved):
LGTM

orion/handlers/grpc/grpc.go
```diff
@@ -31,10 +33,20 @@ type grpcHandler struct {
...
-		)
+		var opts []grpc.ServerOption
+		if g.grpcInterceptor() != nil {
+			opts = append(opts, grpc.UnaryInterceptor(g.grpcInterceptor()))
+		}
+		if g.grpcStreamInterceptor() != nil {
+			opts = append(opts, grpc.StreamInterceptor(g.grpcStreamInterceptor()))
+		}
+		if g.config.CustomCodec != nil {
+			opts = append(opts, grpc.CustomCodec(*g.config.CustomCodec))
```
benjaminheng, line 44: `CustomCodec` is deprecated: https://pkg.go.dev/google.golang.org/grpc#CustomCodec

benjaminheng (changes requested):
.

benjaminheng (approved):

NirbhayK (dismissed):
LGTM

orion/types.go
```diff
@@ -37,6 +38,10 @@ type Server interface {
 	GetConfig() map[string]interface{}
 	//AddInitializers adds the initializers to orion server
 	AddInitializers(ins ...Initializer)
+	//Add custom grpc handler to the orion server
+	AddGrpcUnknownHandler(handler grpc.StreamHandler)
```
ankurs, line 42: is there another approach we can use than to expand the interface, maybe an initialiser ?

GaneshArshavin: hey @ankurs  , Any particular reason why we should not initialiser over expanding interface ?
The current initializers have 
Hystrix
Zpkin
prometheous

but however the interface  method seems to be expanding core server options , i see `unknownHandler` as an extension of the server itself ,than adding it is an initializer . What do you think ?

hothero: +1 on Ankurs' points. To me, the interface describes the Orion server's abilities. However, the server could support single or multiple protocols (e.g. HTTP, gRPC, FTP, etc.) as long as your struct implements the interface.

Currently, your changes here are too protocol-specific in the function naming. Besides that, for engineers, you also need to know if you call AddGrpcUnknownHandler you also need to restart the server to take it effective. How about Initializer? I think it's more straightforward to understand. In addition, Initializer interface has init() and Reinit() functions, that would help GRPCUnknownHandler and HTTPNotFoundHandler more extensible in the future as well.

In short, I would like to see if it's possible to assign the new handlers via initializers.

benjaminheng: It'll be good to explore if Orion can already support unknown gRPC handlers via an initializer. Nothing comes to mind to me right now though, not sure if it's possible.

---

For the unknown HTTP handler, I've posted a comment in one of the RFCs that specify a need for this PR. I'll summarize it here. Orion already supports adding a HTTP handler for unknown routes. Since Orion uses gorilla/mux for route parsing, we can use regex for it.

The following `ORION:URL` directive will capture all unknown routes:

```protobuf
service MyService {
    // ORION:URL: GET/PUT/POST/PATCH/DELETE /{route:[a-zA-Z0-9-_/.]+}
    rpc PassthroughCatchAll(PassthroughRequest) returns (PassthroughResponse);
}
```

A custom encoder can then be used to parse the HTTP metadata into the protobuf message:

```go
func RegisterOptionals(svr orion.Server) {
	proto.RegisterMyServiceDefaultEncoder(svr, defaultRequestEncoder)
	proto.RegisterMyjServiceDefaultDecoder(svr, defaultResponseDecoder)
}

func defaultRequestEncoder(req *http.Request, reqObject interface{}) error {
	// Parse HTTP metadata into the PassthroughRequest protobuf message
	return nil
}

func defaultResponseDecoder(ctx context.Context, w http.ResponseWriter, encodeError, endpointError error, respObject interface{}) {
	if v, ok := respObject.(*proto.PassthroughResponse); ok {
		// Translate PassthroughResponse to HTTP response
		writePassthroughResponse(w, v)
		return
	}
	return
}
```

Implement the handler for unknown HTTP routes:

```go
func (s *service) PassthroughCatchAll(ctx context.Context, req *proto.PassthroughRequest) (*proto.PassthroughResponse, error) {
	// Implement handler for unknown HTTP routes
	return nil, nil
}
```

In fact, this has been implemented by one of our gateways for 2-3 years now, so it's been battle-tested and all of the above functions being referenced can be copied from the service.

hothero: @vikrammais Ben's approach looks good to me. Could you give it a try? otherwise, you can share another PR with functional options pattern.

hothero (changes requested):
per discussions in comments, I'd prefer not having protocol-specific functions in the generic server interface.

orion/types.go
benjaminheng: An alternative is to use Orion's support for [custom HTTP handlers](https://github.com/carousell/Orion/blob/21ea5fa8c92722cf817ad8af8357a71578e90eb3/orion/handlers/types.go#L27-L30) in conjunction with a catch-all route regex.

```protobuf
service UploadSvcService{
    // ORION:URL: GET/PUT/POST/PATCH/DELETE /{route:[a-zA-Z0-9-_/.]+}
    rpc PassthroughCatchAll(PassthroughRequest) returns (PassthroughResponse);
}
```

```go
func (s *svc) GetHTTPHandler(method string) orion.HTTPHandler {
	if method == "passthroughcatchall" { // RPC name, lowercased
		return func(resp http.ResponseWriter, req *http.Request) bool {
			// Implement handler
		}
	}
}
```

This is even simpler than the previous approach, and you get access to the raw incoming `*http.Request` struct.

vikrammais: @hothero  @benjaminheng  for http, as suggested Custom HTTP Handlers is a good option to handle the `not found` http requests. 
But for grpc, couldn't find a similar approach as [UnknownServiceHandler](https://github.com/grpc/grpc-go/blob/master/server.go#L448) 
```
// UnknownServiceHandler returns a ServerOption that allows for adding a custom
// unknown service handler. The provided method is a bidi-streaming RPC service
// handler that will be invoked instead of returning the "unimplemented" gRPC
// error whenever a request is received for an unregistered service or method.
// The handling function and stream interceptor (if set) have full access to
// the ServerStream, including its Context.
func UnknownServiceHandler(streamHandler StreamHandler) ServerOption {
	return newFuncServerOption(func(o *serverOptions) {
		o.unknownStreamDesc = &StreamDesc{
			StreamName: "unknown_service_handler",
			Handler:    streamHandler,
			// We need to assume that the users of the streamHandler will want to use both.
			ClientStreams: true,
			ServerStreams: true,
		}
	})
}
```
expects a StreamHandler to be passed and [set](https://github.com/grpc/grpc-go/blob/a002994200f31f5b3e985e6b2cd2f12cb8b81389/test/end2end_test.go#L601).
```
if te.unknownHandler != nil {
		sopts = append(sopts, grpc.UnknownServiceHandler(te.unknownHandler))
}
```
This can be achieved in two ways
1. [Appending](https://github.com/nhatthm/grpcmock/blob/9dd17902106912d135fca12489b389fdc2086400/server_options.go#L130) to the server options
```
func UnknownServiceHandler(streamHandler grpc.StreamHandler) ServerOption {
	return func(s *Server) {
		s.serverOpts = append(s.serverOpts, grpc.UnknownServiceHandler(streamHandler))
	}
}
```
2. [Initialising](https://github.com/carousell/Orion/compare/test-handlers#diff-1198f55f8eca3ff12264d38ec6018942f91a2a9d3e15bc249963a7a7e859b147R509) server with handler option

But 1 doesn't work for Orion as it does not expose server options. Approach 2 will work without extending the generic server interface and is extendable for any protocol

Thoughts?

hothero: @vikrammais Ideally, I would like to see if it's possible to extend our Orion core to fully support the protobuf annotation. However, I don't have a strong opinion to have that right away.

I think we've discussed the 2. (initialising server with handler option) offline. The structure looks good to me, and it will not make any breaking change affecting all Orion users. Can see if @ankurs and @benjaminheng have any suggestion or idea :) 

btw, I have a few suggestions for the "2" changes, I will leave comments once you initiated the PR.

orion/core.go
```diff
@@ -499,3 +502,39 @@ func GetDefaultServerWithConfig(config Config) Server {
 		config: config,
 	}
 }
+
+func GetDefaultServerWithHandlerOptions(name string, opts ...DefaultServerOption) Server {
```
hothero, line 506: do you think we can just extend "GetDefaultServer"? `opts ...DefaultServerOption` is an optional arguments so it won't cause any breaking change.

orion/core.go
```diff
@@ -68,10 +68,11 @@ type middlewareInfo struct {
...
 //DefaultServerImpl provides a default implementation of orion.Server this can be embedded in custom orion.Server implementations
 type DefaultServerImpl struct {
-	config Config
-	mu     sync.Mutex
-	wg     sync.WaitGroup
-	inited bool
+	config                    Config
+	mu                        sync.Mutex
+	wg                        sync.WaitGroup
+	grpcUnknownServiceHandler *grpc.StreamHandler
```
benjaminheng, line 74:
```suggestion
	grpcUnknownServiceHandler grpc.StreamHandler
```

orion/core.go
```diff
@@ -499,3 +502,39 @@ func GetDefaultServerWithConfig(config Config) Server {
...
+
+type DefaultServerOption interface {
+	apply(*DefaultServerImpl)
+}
+
+// WithGrpcUnknownHandler returns a DefaultServerOption which sets
+// UnknownServiceHandler option in grpc server
+func WithGrpcUnknownHandler(grpcUnknownServiceHandler grpc.StreamHandler) DefaultServerOption {
+	return newFuncDefaultServerOption(func(h *DefaultServerImpl) {
+		h.grpcUnknownServiceHandler = &grpcUnknownServiceHandler
```
benjaminheng, line 524:
```suggestion
		h.grpcUnknownServiceHandler = grpcUnknownServiceHandler
```

orion/handlers/grpc/grpc.go
```diff
@@ -31,10 +33,17 @@ type grpcHandler struct {
...
-		)
+		var opts []grpc.ServerOption
+		if g.grpcInterceptor() != nil {
+			opts = append(opts, grpc.UnaryInterceptor(g.grpcInterceptor()))
+		}
+		if g.grpcStreamInterceptor() != nil {
+			opts = append(opts, grpc.StreamInterceptor(g.grpcStreamInterceptor()))
+		}
+		if g.config.UnknownServiceHandler != nil {
+			opts = append(opts, grpc.UnknownServiceHandler(*g.config.UnknownServiceHandler))
```
benjaminheng, line 44:
```suggestion
			opts = append(opts, grpc.UnknownServiceHandler(g.config.UnknownServiceHandler))
```

benjaminheng (changes requested):
`DefaultServerImpl.grpcUnknownServiceHandler` doesn't have to be a pointer to a function. 

Currently we're
1. taking in a non-pointer function: `WithGrpcUnknownHandler(f grpc.StreamHandler)`
2. storing it as a pointer: `h.grpcUnknownServiceHandler = &grpcUnknownServiceHandler`
3. then taking the value again when applying the option: `opts = append(opts, grpc.UnknownServiceHandler(*g.config.UnknownServiceHandler))`

My guess is that you're storing a pointer so that you can check if `g.config.UnknownServiceHandler != nil` when applying the option. However function types are already nillable, so you don't need to use pointers for nillability.

Other than this I'm ok with using functional options pattern for this.

hothero (dismissed):
LGTM

benjaminheng (dismissed):

benjaminheng:
@vikrammais Hey Vikram, I notice this PR hasn't been merged despite being approved. Is it still required?

achichen:
@vikrammais  @benjaminheng @hothero @NirbhayK As this PR has been approved. I am merging this PR. I've also solved the conflict in core.go.

achichen (approved):

PR #172, opened 2022-03-12
Support tool installation after Go 1.17

cs-lexliu:
### Situation:
Starting in Go 1.17, installing executables with `go get` is deprecated. `go install` may be used instead. And go get is no longer supported outside a module in Go 1.18. ([Source](https://go.dev/doc/go-get-install-deprecation))

Which means Orion is no longer available in Go 1.18. We will stuck with older version of Go in the future.

Install `protoc-gen-orion` with different Go version.

||Go 1.16|Go 1.17|Go 1.18|
|-|-|-|-|
|go get|Passed|Warned|Failed|
|go install|Failed|Failed|Failed|

> go get -u github.com/carousell/Orion/protoc-gen-orion
> go install github.com/carousell/Orion/protoc-gen-orion@latest

### Target:
Allow `protoc-gen-orion` to be installed by `go install`.

### Action:
Remove replace directives from `go.mod`.

### Result:
We can install `protoc-gen-go` with `go install` command in Go 1.16, 1.17, and 1.18.

Install `protoc-gen-orion` with different Go version.

||Go 1.16|Go 1.17|Go 1.18|
|-|-|-|-|
|go get|Passed|Warned|Failed|
|go install|Passed|Passed|Passed|

> go get -u github.com/carousell/Orion/protoc-gen-orion@afd0407a
> go install github.com/carousell/Orion/protoc-gen-orion@afd0407a
> afd0407a = this commit


hothero:
I think I kinda get the purpose of having "replace" statement in Orion, it should be to help Orion's consumers to no need to add replace descriptive, instead, we have the "forcing version" in the root-level Orion project.

I believe we still need "replace" anyway to deal with breaking changes and circular dependencies, so we need to figure out how to force version in Golang 1.18. It looks like we have to fork the package? https://www.sobyte.net/post/2022-01/go-multi-module/

achichen:
@cs-lexliu  Will it be resolved with less impact if we add go.mod under `Orion/protoc-gen-orion` to make it a standalone module?

cs-lexliu:
> @cs-lexliu Will it be resolved with less impact if we add go.mod under `Orion/protoc-gen-orion` to make it a standalone module?

That's a great idea. Let me try.

cs-lexliu:
@achichen It works. Thanks. 

@hothero In the final solution, we don't remove the replacement from root go.mod. I think the concern of breaking change is solved.

Please check it again.

hothero (approved):
new approach looks good to me!

achichen (approved):
LGTM

PR #174, opened 2022-04-13
Deprecate trace function in notifier package

praveensanap:
This PR deprecates trace-id functionality from notifier package.

notifier package relies on "trace" set in the span baggage of the parent span based on opentracing spec.

With release of Metadata passing functionality, we want metadata to carry trace information and not the spancontext baggage.
https://github.com/carousell/Orion/pull/149

We recommend forwarding trace information using  `grpc_opentracing.UnaryClientInterceptor`
 from `github.com/grpc-ecosystem/go-grpc-middleware` module.


praveensanap:
closing for now. because this is not blocking out Unified log, trace sampling initiative

PR #175, opened 2022-04-18
Remove missing dependency `micro/protobuf`

cs-lexliu:
The package `micro/protobuf` has been removed from github.
Go return error when executing `go mod tidy`, because the repository is unreachable.
Replace the package with `protobuf/protoc-gen-go/generator` which has the same functionality.


Makefile
```diff
@@ -8,7 +8,10 @@ all: clean vet test build
 
 cleanall: clean dockerclean
 
-ci: clean vet bench build
+ci: mod clean vet bench build
+
+mod:
```
cs-lexliu, line 13: When Travis execute `go clean ./...`, it failed because there are some packages missing on the device.
Execute `go mod tidy` before clean action to arrangement and download the dependency packages.


https://app.travis-ci.com/github/carousell/Orion/jobs/567329539

benjaminheng: We shouldn't do `go mod tidy` as part of the CI process, because that hides problems. The changes made `go mod tidy` normally needs to be committed into the repo; doing so in CI might "fix" the issue, but the repo still remains in an invalid state.

There are actually a few deeper problems that you uncovered:

1. We should be allowing failures from master. In .travis.yml we configured it to allow failures on the `master` branch, but we're running CI builds against `tip`. We should be running builds against `master` instead.
2. The `go vet` error you see in the `tip` build is because `golang.org/x/crypto/` needs to be updated for Go 1.18+.
3. Not pictured in your failing build, but `golang.org/x/sys` also needs to be updated for Go 1.18+ on macos: https://github.com/golang/go/issues/49219

I've made the changes in this branch: https://github.com/carousell/Orion/compare/remove-missing-dependency...remove-missing-dependency-2

Feel free to merge my branch into yours

cs-lexliu: Thanks. I will merge your branch later.

praveensanap (approved):
LGTM

PR #176, opened 2022-04-18
Fix travis ci error

cs-lexliu:
We have an error when travis building code on tip env with go1.19.
Run `go mod tidy` for download missing packages on the env.
Add more go version for building the code.


cs-lexliu:
For experiment.

PR #177, opened 2022-04-20
Fix passing trace metadata

praveensanap:
This PR fixes metadata forwarding issue for trace headers present in `DefaultClientInterceptors`.

**Current behaviour**
**Client Side**
Outgoing grpc calls from Orion services are recommended to user `DefaultClientInterceptors`.
Among other, these interceptors are `GRPCClientInterceptor` and `ForwardMetadataInterceptor`, in that order.

`GRPCClientInterceptor`  is a wrapper around [grpc_opentracing.UnaryClientInterceptor](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/tracing/opentracing/client_interceptors.go#L21)
It starts a client span and adds the tracing headers to [outgoing metadata](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/68c8cdc5fd8157730d9a930d9c8b7166cfae31ac/tracing/opentracing/client_interceptors.go#L133) using `metadata.NewOutgoingContext`

`ForwardMetadataInterceptor` reads the incoming metadata from context and appends to outgoing metadata context using `metadata.AppendToOutgoingContext`

The result of these two interceptor can be illustrated as following:
Suppose ServiceA -> ServiceB, the outgoing metadata in this grpc call looks like:
```
{
    "traceid" : ["<trace id from new grpc client span>", "<trace id in the incoming metadata>"],
    "spanid" : ["<span id from new grpc client span>", "<span id in the incoming metadata>"],
// .... other trace headers.

}
```
Note: the trace header key will depend on the propagation mechanism configured in tracer initialisation.

**Server side**
The Orion server receiving these header will read them in the [grpc_opentracing.UnaryServerInterceptor](https://github.com/carousell/Orion/blob/336d176e1875f0feca35c45f6d9e135e0d83c621/interceptors/interceptors.go#L34)

This interceptors reads trace headers from metadata, create a parent span context from them and sets it to a newly created service side span.

This logic in this interceptors read the last value for a particular key. So in our example `<trace id in the incoming metadata>` and `<span id in the incoming metadata>` will be used to create the parent span context.

This is undesirable because a wrong parent context is created at the server side. The parent context should be of the client span created by `grpc_opentracing.UnaryClientInterceptor`.

**Solution**
This PR fixes the order of client interceptors to `ForwardMetadataInterceptor` before `GRPCClientInterceptor`.

`ForwardMetadataInterceptor` appends the incoming metadata to outgoing metadata as is.
But `GRPCClientInterceptor` overwrites (since it uses `metadata.NewOutgoingContext`) the trace metadata with the current span info. 

In above example this will be received by ServiceB server interceptor. 

```
{
    "traceid" : ["<trace id from new grpc client span>],
    "spanid" : ["<span id from new grpc client span>],
// .... other trace headers.

}
```

This is desirable since the server interceptors only need to know the span info of it's caller.


interceptors/interceptors.go
```diff
@@ -43,10 +43,10 @@ func DefaultInterceptors() []grpc.UnaryServerInterceptor {
 func DefaultClientInterceptors(address string) []grpc.UnaryClientInterceptor {
 	return []grpc.UnaryClientInterceptor{
 		grpc_retry.UnaryClientInterceptor(),
-		GRPCClientInterceptor(),
 		NewRelicClientInterceptor(address),
 		HystrixClientInterceptor(),
 		ForwardMetadataInterceptor(),
+		GRPCClientInterceptor(),
```
praveensanap, line 49: Order of `GRPCClientInterceptor`  `ForwardMetadataInterceptor` important.
Because the trace headers propagated to the caller service should be of the current span, not the upstream span present in the incoming metadata.

achichen (approved):

edwardsamuel (approved):

PR #179, opened 2022-06-20
Adding sonar.properties and .golangci.yml template files

priyankshah217:
- Adding sonarqube properties files as a part of service creation 
- Adding .golangci.yml as a part of service creation
- Adding phase to add golangci dependency


achichen:
Close this in favor of Orion-Builder one

PR #182, opened 2022-07-05
add a unit test to verify the interceptors execution order is normal

hothero:
Orion's [chainUnaryServer](http://github.com/carousell/Orion/blob/c54848a93c8e634429d4b150879a061e93a9e9e8/orion/handlers/utils.go#L22-L22) executes the interceptors in a normal order while grpc's ChainUnaryInterceptors do it in a reverse order. This added unit test is to ensure the Orion's chainUnaryServer executes interceptors as expected. And, Orion's users can have high confidence at understanding it.

### DIFF
--- a/orion/handlers/utils_test.go
+++ b/orion/handlers/utils_test.go
@@ -21,7 +21,7 @@ func (s *service) GetInterceptors() []grpc.UnaryServerInterceptor {
 	return s.interceptors
 }
 
-func TestGetInterceptors(t *testing.T) {
+func TestGetInterceptorsWithMethodMiddlewares(t *testing.T) {
 	// this test currently verifies the Interceptors is executed at the same sequence that you passed in
 	// gRPC's normal chainUnaryInterceptor is executing interceptor in a reverse order.
 	// but Orion's chainUnaryServer ensure the interceptor execution order is normal (ensure this)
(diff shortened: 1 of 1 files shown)


orion/handlers/utils_test.go
```diff
@@ -0,0 +1,55 @@
...
+			return handler(ctx, req)
+		},
+		func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
+			sequenceArr = append(sequenceArr, 3)
+			return handler(ctx, req)
+		},
+	}
+	svc := NewMockService(interceptors...)
+
+	serverInterceptor := GetInterceptors(svc, CommonConfig{NoDefaultInterceptors: true})
```
achichen, line 47: It seems that `GetInterceptors()` is not directly or indirectly used in Orion's code base. While it's `GetInterceptorsWithMethodMiddlewares()` being used by grpc and http handler to get the interceptors to use.

If the intent is to ensure Orion calls interceptors in the normal order, should we test `GetInterceptorsWithMethodMiddlewares()` instead? Or test the private/unexposed function `chainUnaryServer()` directly?

hothero: good call, I changed to test GetInterceptorsWithMethodMiddlewares

achichen (approved):

benjaminheng (approved):

PR #185, opened 2022-08-03
[RE-5863] add test cases for fields

jalacardio:
## Description

Context

## Checklist
- [ ] Test on local
- [ ] Test on stage
- [ ] run `go mod tidy` and `go mod vendor`
- [ ] change module's version back to master 

## Impacted Endpoints (Optional)

### DIFF
--- a/utils/log/loggers/test/parallelism_test.go
+++ b/utils/log/loggers/test/parallelism_test.go
@@ -1,0 +1,107 @@
+//go test -race
+package loggers_test
+
+import (
+	"context"
+	"fmt"
+	"math/rand"
+	"sync"
+	"testing"
+	"time"
+
+	s "github.com/carousell/Orion/utils/log/loggers"
+	"github.com/stretchr/testify/assert"
+)
+
+const readWorkerCount = 5
+const writeWorkerCount = 5
+
+func readWorker(idx int, ctx context.Context) {
+	lf := s.FromContext(ctx)
+	// simulate reading task
+	time.Sleep(time.Millisecond * 250)
+	fmt.Printf("Reader %d read from logfields %+v\n", idx, lf)
+}
+
+func writeWorker(idx int, ctx context.Context) context.Context {
+	key := fmt.Sprintf("key%d", idx)
+	val := fmt.Sprintf("val%d", rand.Intn(10000))
+	ctx = s.AddToLogContext(ctx, key, val)
+	time.Sleep(time.Millisecond * 250)
+	fmt.Printf("Writer %d wrote %s:%s\n", idx, key, val)
+	return ctx
+}
+
+func TestParallelRead(t *testing.T) {
+	// LogContext init, non-paralel
+	ctx := context.Background()
+	ctx = s.AddToLogContext(ctx, "k1", "v1")
+	ctx = s.AddToLogContext(ctx, "k2", "v2")
+
+	var wg sync.WaitGroup
+	for i := 1; i <= readWorkerCount; i++ {
+		wg.Add(1)
+		go func(j int) {
+			defer wg.Done()
+			readWorker(j, ctx)
+		}(i)
+	}
+	wg.Wait()
+}
+
+func TestParallelWrite(t *testing.T) {
+	ctx := context.Background()
+	ctx = s.AddToLogContext(ctx, "test-key", "test-value")
+
+	var wg sync.WaitGroup
+	for i := 1; i <= writeWorkerCount; i++ {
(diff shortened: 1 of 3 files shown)


utils/log/loggers/test/parallelism_test.go
```diff
@@ -0,0 +1,103 @@
...
+	var wgWrite sync.WaitGroup
+	for i := 1; i <= writeWorkerCount; i++ {
+		wgWrite.Add(1)
+		go func(j int) {
+			defer wgWrite.Done()
+			writeWorker(j, ctx)
+		}(i)
+	}
+	wgRead.Wait()
+	wgWrite.Wait()
```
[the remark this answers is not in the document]
jalacardio: okay! changed.

utils/log/loggers/fields.go
```diff
@@ -12,6 +13,10 @@ var (
 
 //LogFields contains all fields that have to be added to logs
 type LogFields map[string]interface{}
+type ProtectedLogFields struct {
+	Content LogFields
+	mtx     sync.RWMutex
```
achichen, line 18: is testing from `loggers_test` package the only reason to expose this struct? if so, what's the downside to write test in the same package?

the same for `Content`, if we write test from the same package, there is no need to expose this field and this keeps your encapsulation better.

utils/log/loggers/fields.go
```diff
@@ -46,8 +54,26 @@ func FromContext(ctx context.Context) LogFields {
...
+		if plf, ok := h.(*ProtectedLogFields); ok {
+			plf.mtx.RLock()
+			defer plf.mtx.RUnlock()
+			content := make(LogFields)
+			for k, v := range plf.Content {
+				content[k] = v
+			}
+			return content
+		}
+	}
```
achichen, line 66: to avoid duplicated logic, would it be better to wrap this block using `fromContext`?

```golang
if plf := fromContext(ctx); plf != nil {
    plf.mtx.RLock()
    defer plf.mtx.RUnlock()
    content := make(LogFields)
    for k, v := range plf.Content {
        content[k] = v
    }
    return content
}
```

utils/log/loggers/fields.go
```diff
@@ -28,14 +33,17 @@ func (o LogFields) Del(key string) {
...
+		data = fromContext(ctx)
 	}
 	m := ctx.Value(contextKey)
-	if data, ok := m.(LogFields); ok {
-		data.Add(key, value)
+	if data, ok := m.(*ProtectedLogFields); ok {
+		data.mtx.Lock()
+		defer data.mtx.Unlock()
+		data.Content.Add(key, value)
 	}
```
achichen, line 47: can just use `data`, without need to take `m` out from context again.  the original code seems redundant, can it be simplified to following code?

```golang
data := fromContext(ctx)
if data == nil {
    data = &ProtectedLogFields{Content: make(LogFields)}
    ctx = context.WithValue(ctx, contextKey, data)
}
data.mtx.Lock()
defer data.mtx.Unlock()
data.Content.Add(key, value)
```

utils/log/loggers/fields.go
jalacardio: You're right, there's no need to expose these structs, changed and thanks.

utils/log/loggers/fields.go
jalacardio: Great idea, changed, thanks!

utils/log/loggers/fields.go
jalacardio: This is brilliant, I think the reason it was previously written like that was due to it needs to fetch latest copy after save, since now we are using pointer directly, this can be skipped. Great catch and thanks!

ankurs:
shouldn't you just change logfields to sync.Map instead ?

something along the lines of - https://github.com/go-coldbrew/log/pull/5/files

jalacardio:
That is in fact a more concise way to dealing with it, thanks for the info and will look into it.

utils/log/loggers/fields.go
```diff
@@ -12,6 +13,10 @@ var (
 
 //LogFields contains all fields that have to be added to logs
 type LogFields map[string]interface{}
+type protectedLogFields struct {
+	content LogFields
+	mtx     sync.RWMutex
```
achichen, line 18: As @ankurs suggested, `sync.Map` is indeed a better option.

According to https://pkg.go.dev/sync#Map, it optimizes for our use case:
> when the entry for a given key is only ever written once but read many times, as in caches that only grow

Issue #186, opened 2022-09-07
Allow *.orion.pb.go files to be generated into a different package from the generated protobufs


benjaminheng:
Currently the generated *.orion.pb.go file needs to be placed as a sibling to the generated protobufs, because the orion file references an unexported variable:

https://github.com/carousell/Orion/blob/ec021a79aefef548f78fd349884cb3f50b8db9ed/protoc-gen-orion/orion.go#L262

In later versions of protoc-gen-go-grpc, this variable is now exported. https://github.com/carousell/Orion/issues/171 is a related issue to allow protoc-gen-orion to use the exported variable.

The variable being exported now also means that it should be possible to place the *.orion.pb.go file in a separate package. In Carousell we're considering having a Go module for all generated protobufs, and another module for all generated orion *.orion.pb.go files.

Ideally the module for generated protobufs will only contain two dependencies:

```
github.com/golang/protobuf
google.golang.org/grpc
```

Having the orion files as a sibling will require `github.com/carousell/Orion` to be added as another dependency. The orion dependency is too heavy for us to include. Not every service that consumes our generated protobufs will need Orion. 

Thus ideally the orion file will live in a separate package. Example file structure:

```
gen/
  go/
    serviceA/serviceA.pb.go
    serviceB/serviceB.pb.go
    go.mod
  orion/
    serviceA/serviceB.orion.pb.go
    go.mod
```

I have some ideas of how to extend protoc-gen-orion to support this. I'll add more comments to the issue later.

cs-lexliu:
Before we use the latest version of proto-gen-go to get the exported service desc.
We have to modify the `go_plugin_install` to support higher proto-gen-go version.
The current proto-gen-go we used is v1.3.1, and this is getting from [github.com/golang/protobuf/protoc-gen-go](https://pkg.go.dev/github.com/golang/protobuf/protoc-gen-go?tab=versions).
The higher version (> v1.5.2) of proto-gen-go has been change to [google.golang.org/protobuf/cmd/protoc-gen-go](https://pkg.go.dev/google.golang.org/protobuf/cmd/protoc-gen-go?tab=versions).

We also have to introduce [google.golang.org/grpc/cmd/protoc-gen-go-grpc](https://pkg.go.dev/google.golang.org/grpc/cmd/protoc-gen-go-grpc?tab=versions) into our protogen tool. Because it's separated from protoc-gen-go after v1.20.0.

cs-lexliu:
After protogen supports higher version.
[...]

PR #189, opened 2022-09-21
Allow *.orion.pb.go files to access the exported-service-desc from *.pb.go

cs-lexliu:
Add two execution parameters.
`standalone-mode`: Allow you to place *.orion.pb.go and *.pb.go into different folder.
`exported-service-desc`: Allow you to use the higher version (v1.20.0) of protoc-gen-go. Read go_package option from protobufs files.
In standalone mode, the external import path is getting from the go_package option.

Check [protoc-gen-orion/testprotos/Makefile](https://github.com/carousell/Orion/blob/26e9b2d59f6245c4dd77e449c65fb6e1e9d08e51/protoc-gen-orion/testprotos/Makefile) for the usage.

#186


protoc-gen-orion/orion.go
```diff
@@ -96,13 +97,21 @@ type orionMiddleware struct {
 	Names      string
 }
 
+type ProtocParams struct {
+	StandAloneMode      bool
```
benjaminheng, line 101: change to `StandaloneMode`, because standalone is a single word

benjaminheng (dismissed):

protoc-gen-orion/orion.go
cs-lexliu: fixed.

benjaminheng (approved):

protoc-gen-orion/testprotos/standalone_mode/example.go
```diff
@@ -0,0 +1,5 @@
+package example
+
+import "google.golang.org/grpc"
+
+var ExampleService_ServiceDesc grpc.ServiceDesc
```
praveensanap, line 5: @cs-lexliu Can you variable generated from protoc-gen-go
Which version of protoc-gen-go should be used to generate it?
because current version would generate _ExampleService_ServiceDesc.

Do we have to bump the version in shared-proto scripts?

praveensanap (approved):

protoc-gen-orion/testprotos/standalone_mode/example.go
cs-lexliu: Let me create another PR to describe how to use the tool and supported grpc version for the tool

PR #193, opened 2022-12-06
[BSX-265] improve grpc-logging

eric-lin-git:
https://carousell.atlassian.net/browse/BSX-265

old log format
```
{"@timestamp":"2022-12-06T08:43:44.955187131Z","caller":"interceptors/unary_interceptors.go:38","error":null,"grpcMethod":"/warrantycorepb.WarrantyCoreService/GetWarrantyOptionGroup","level":"info","method":"/warrantycorepb.WarrantyCoreService/GetWarrantyOptionGroup","took":"7.902368ms","trace":"1802ae6a-7542-11ed-926a-8ac647f8a923","trace_id":"1d4a11bb0c05c2ba","transport":"gRPC"}
```

new log format
```
{"@timestamp":"2022-12-06T08:34:51.817602277Z","caller":"interceptors/unary_interceptors.go:74","grpcMethod":"/warrantycorepb.WarrantyCoreService/GetWarrantyOptionGroup","level":"info","req":"{\"ccid\":\"2799\",\"listing_type\":[1],\"warranty_option_length_unit\":1}","resp":"{\"warranty_option_group\":{\"ccid\":\"2799\",\"options\":[{\"id\":\"a85fef11-1ff9-46d6-ad63-ce76735c7bf4\",\"currency\":\"MYR\",\"price\":\"0\",\"length\":1},{\"id\":\"e63ae5d0-848c-4f3a-bca5-df26152d328e\",\"currency\":\"MYR\",\"price\":\"49\",\"length\":6},{\"id\":\"8f1aa172-d126-4060-8665-4f7d6ed84347\",\"currency\":\"MYR\",\"price\":\"88\",\"length\":12,\"is_best\":true},{\"id\":\"f0c319dc-2a82-47b6-a2ab-8648cb93d6ea\",\"currency\":\"MYR\",\"price\":\"180\",\"length\":24}],\"best_option_id\":\"8f1aa172-d126-4060-8665-4f7d6ed84347\"}}","took":"4.534181ms","trace":"da3ccd00-7540-11ed-a081-aaa3ee5b0697","trace_id":"79d71ec1fa946f59","transport":"gRPC"}
```

What we have changed in this PR:
1. add the logging for req and resp, default not print
2. remove to print `"error":null` if no error
3. remove to print `method` as it is already print in `grpcMethod`

### DIFF
--- a/interceptors/unary_interceptors.go
+++ b/interceptors/unary_interceptors.go
@@ -29,6 +29,7 @@ func DebugLoggingInterceptor() grpc.UnaryServerInterceptor {
 	}
 }
 
+// Deprecated: use GRPCLoggingInterceptor instead
 // ResponseTimeLoggingInterceptor logs response time for each request on server
 func ResponseTimeLoggingInterceptor() grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
--- a/interceptors/unary_interceptors.go
+++ b/interceptors/unary_interceptors.go
@@ -71,9 +71,9 @@ func GRPCLoggingInterceptor(enableRequestLog, enableResponseLog bool) grpc.Unary
 				}
 
 				if hasErr {
-					log.Warn(ctx, args)
+					log.Warn(ctx, args...)
 				} else {
-					log.Info(ctx, args)
+					log.Info(ctx, args...)
 				}
 
 			}(time.Now())
--- a/interceptors/unary_interceptors.go
+++ b/interceptors/unary_interceptors.go
@@ -57,11 +57,11 @@ func GRPCLoggingInterceptor(enableRequestLog, enableResponseLog bool) grpc.Unary
 				enableResponseLog = enableResponseLog || hasErr
 
 				if enableRequestLog {
-					args = append(args, "req", req)
+					args = append(args, "req", utils.ToJSONString(req))
 				}
 
 				if enableResponseLog {
-					args = append(args, "resp", resp)
+					args = append(args, "resp", utils.ToJSONString(resp))
 				}
 
 				if hasErr {
--- a/interceptors/unary_interceptors.go
+++ b/interceptors/unary_interceptors.go
@@ -49,8 +49,6 @@ func GRPCLoggingInterceptor(enableRequestLog, enableResponseLog bool) grpc.Unary
 		if !modifiers.IsHTTPRequest(ctx) {
 			defer func(begin time.Time) {
 				args := []interface{}{
-					"method",
-					info.FullMethod,
 					"took",
 					time.Since(begin),
 				}
--- a/interceptors/unary_interceptors.go
+++ b/interceptors/unary_interceptors.go
@@ -19,7 +19,7 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
-//DebugLoggingInterceptor is the interceptor that logs all request/response from a handler
+// DebugLoggingInterceptor is the interceptor that logs all request/response from a handler
 func DebugLoggingInterceptor() grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 		fmt.Println(info, "requst", req)
@@ -29,7 +29,7 @@ func DebugLoggingInterceptor() grpc.UnaryServerInterceptor {
 	}
 }
 
-//ResponseTimeLoggingInterceptor logs response time for each request on server
+// ResponseTimeLoggingInterceptor logs response time for each request on server
 func ResponseTimeLoggingInterceptor() grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
 		// dont log for HTTP request, let HTTP Handler manage it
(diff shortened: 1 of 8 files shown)


benjaminheng:
Regarding removing the `method` field, @vipulsodha do you know if our logging service reads from the `method` field or `grpcMethod` field when ingesting logs? I'm assuming it reads from the latter instead.

orion/handlers/utils.go
```diff
@@ -117,6 +117,8 @@ func getInterceptors(svc interface{}, config CommonConfig, middlewares []string)
 		opts = append(opts, interceptors.DefaultInterceptors()...)
 	}
 
+	opts = append(opts, interceptors.GRPCLoggingInterceptor(config.EnableGRPCRequestLog, config.EnableGRPCResponseLog))
```
praveensanap, line 120: Is there a reason `GRPCLoggingInterceptor` is adding this way and not in `DefaultInterceptors`?

praveensanap:
@benjaminheng 
We ingest both "method" and "grpcMethod" to gcp-logs.
https://github.com/carousell/log-svc-worker/blob/b930b9ed20e953ced25c8eb8cf6e0fffe73f5228/log-svc-worker/service/consumer/ingestors/gcp_logging_ingestor.go#L38

I prefer keeping grpcMethod.

praveensanap:
@eric-lin-git 
On 2. remove to print "error":null if no error

I use this field to filter out non error logs by `grep -v "error\":null"`  which I understand is workaround to get error logs.
I'm afraid developers would have to change their habits and find new ways to checking logs.

although removing null fields make sense to save disk space.

orion/config.go
```diff
@@ -55,6 +56,13 @@ type Config struct {
 	DisableDefaultInterceptors bool
 	// Receive message Size is used to update the default limit of message that can be received
 	MaxRecvMsgSize int
+
+	GRPCConfig GRPCConfig
+}
+
+type GRPCConfig struct {
```
praveensanap, line 63: GRPCLogConfig is a better name?

utils/utils.go
```diff
@@ -86,11 +88,19 @@ func FinishNRTransaction(ctx context.Context, err error) {
...
 	}
 	return nil
 }
+
+func ToJSONString(input interface{}) string {
+	b, err := json.Marshal(input)
+	if err != nil {
+		return ""
+	}
+	return *(*string)(unsafe.Pointer(&b))
```
praveensanap, line 105: plain `string(b)` doesn't fit the bill here?
trying to understand the usecase of unsafe pointer. since we already know the input ([]byte) and output (string) types.

orion/handlers/utils.go
eric-lin-git: In my opinion having a `DefaultInterceptors` is not a good idea - it breaks the single responsibility principle and from the method name it is difficult to guess what it does, what it's contract

how do you think @praveensanap

eric-lin-git:
> https://github.com/carousell/log-svc-worker/blob/b930b9ed20e953ced25c8eb8cf6e0fffe73f5228/log-svc-worker/service/consumer/ingestors/gcp_logging_ingestor.go#L38

Hi @benjaminheng @praveensanap is there any impact if we remove the `method` in the log?

interceptors/unary_interceptors.go
```diff
@@ -43,7 +44,45 @@ func ResponseTimeLoggingInterceptor() grpc.UnaryServerInterceptor {
...
+		// dont log for HTTP request, let HTTP Handler manage it
+		if !modifiers.IsHTTPRequest(ctx) {
+			defer func(begin time.Time) {
+				args := []interface{}{
+					"took",
+					time.Since(begin),
+				}
+				hasErr := err != nil
+				enableRequestLog = enableRequestLog || hasErr
+				enableResponseLog = enableResponseLog || hasErr
```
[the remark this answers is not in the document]
eric-lin-git: this is a good idea - but if we design like this, probably we have to use `string` type instead of a `enum`

1. actual flag user is in `interceptors` pkg
2. the flag will be passed from `handlers` pkg
3. we cannot import `orion.Config` to `handlers` (otherwise cyclic import and required more refactor)

that means each package we need to define the same enum, which I think is too overkill
I suggest we just keep this for now (print resp/req if err), how do you think?

eric-lin-git:
>

utils/utils.go
eric-lin-git: this is for performance consideration - `string(b)` will allocate a new memory space and copy the data from byte[].
for this log every request/response purpose we can try to avoid it

https://stackoverflow.com/questions/59209493/how-to-use-unsafe-get-a-byte-slice-from-a-string-without-memory-copy

interceptors/unary_interceptors.go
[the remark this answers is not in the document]
benjaminheng: I'm not in favor of changing the default behavior to log the request/response when there's an error.

1. Orion is open sourced, not an internal framework. We need to be mindful about not changing default behavior.
2. Logging request/response on errors is too opinionated. It's not something every service will want. The obvious example is PII data; unilaterally logging the request will risk accidentally logging PII data.

For opinionated changes like this, how about adding a new server interceptor to our `logging` package in go-utils? Your interceptor can use `AddToLogContext` to add the request/response to the log. The package has some other interceptors using this approach, which you can reference. Your services can then explicitly include the new interceptor if they want this behavior.

orion/handlers/utils.go
benjaminheng: 1. The `if !config.DisableDefaultInterceptors` check doesn't disable for your interceptor. Your interceptor is included even if a service is explicitly trying to disable all default interceptors.
2. Specifying all the default interceptors in `DefaultInterceptors` makes it easier to inspect what the default interceptors are. Whereas with this approach, some default interceptors are specified in one place, and some are specified in another place. It's overall less readable.

eric-lin-git:
thanks @benjaminheng @praveensanap ,
moving the implementation to https://github.com/carousell/go-utils/pull/122 to avoid changes on Orion

PR #206, opened 2023-03-07
Moveout utils

achichen:
Sibling PR: https://github.com/carousell/go-utils/pull/125

This PR moves out following independent packages from Orion to go-utils as a standalone modules:
- `orion/utils/executor` -> `go-utils/executor`
- `orion/utils/pubsub` -> `go-utils/pubsub`
- `orion/utils/kafka` -> `go-utils/kafka`
- `orion/utils/worker` -> `go-utils/worker`
- `orion/utils/migrator` -> `go-utils/migrator`
- `orion/utils/httptripper` -> `go-utils/httptripper`

These modules can operate without Orion and should be a drop-in replacement for all use cases.


ankurs:
It looks like the repo we are moving these to is not public, is that expected ?

praveensanap (approved):

achichen:
Hey @ankurs, it's good to hear from you! 

Yes, we are now working on new Orion with better modularization. All the packages in this move can operate independently from Orion without dependency in between. We decide to make them individual modules out of Orion and move into our private repo for now. We might open source them in future if it's useful for general cases. If you see any one are specifically useful for open source, please let me know. 

Btw, do you own https://github.com/go-orion? We are discussing reorganizing source code for new Orion and its eco-systems. One option is to place them as separate repos under `go-orion`, similar to what you do in https://github.com/orgs/go-coldbrew/repositories.

ankurs:
Hey @achichen always here to help, haha

>Yes, we are now working on new Orion with better modularization. All the packages in this move can operate independently from Orion without dependency in between. We decide to make them individual modules out of Orion and move into our private repo for now. We might open source them in future if it's useful for general cases. If you see any one are specifically useful for open source, please let me know.

That's interesting, if you have some goals in mind and would prefer to collaborate on https://github.com/go-coldbrew we can make it happen

>Btw, do you own https://github.com/go-orion? We are discussing reorganizing source code for new Orion and its eco-systems. One option is to place them as separate repos under go-orion, similar to what you do in https://github.com/orgs/go-coldbrew/repositories.
Yes, I do own go-orion, happy to let you guys use that space

uneet77 (approved):

ankurs:
@achichen makes sense, looking at the proposed changes, it feels like you are building [coldbrew](https://github.com/go-coldbrew/), haha

let me know if you guys prefer to collaborate on coldbrew, its fairly stable you can give it a try at https://docs.coldbrew.cloud/

PR #221, opened 2023-12-27
Remove go module replace

achichen:
The two go module replacement was added long time ago and no longer needed. See the line comment for the explanation


go.mod
```diff
@@ -2,11 +2,6 @@ module github.com/carousell/Orion
 
 go 1.17
 
-replace (
-	git.apache.org/thrift.git => github.com/apache/thrift v0.12.0
```
achichen, line 6: As `git.apache.org/thrift.git` is not even in direct or indirect dependency listed in this go.mod, it doesn't effectively affect the result of dependency resolution. The replace is no effect, can be safely removed.

go.mod
```diff
@@ -2,11 +2,6 @@ module github.com/carousell/Orion
...
-	github.com/openzipkin-contrib/zipkin-go-opentracing => github.com/openzipkin-contrib/zipkin-go-opentracing v0.3.5
```
achichen, line 7: `github.com/openzipkin-contrib/zipkin-go-opentracing@v0.3.5` is explicitly listed in this go.mod. no `replace` is needed to override the version

uneet77 (approved):

PR #230, opened 2024-09-18
Read timeouts config for orion server

bhadreswar-ghuku:
We needed to have custom timeout for http handler (default is 10s) for longer api calls in BifrostGateway [[PR](https://github.com/carousell/BifrostGateway/pull/237/files)]. 
It also applies server option. There was no method to do that. Hence adding the method.

### DIFF
--- a/orion/core.go
+++ b/orion/core.go
@@ -513,6 +513,17 @@ func GetDefaultServer(name string, opts ...DefaultServerOption) Server {
 	return server
 }
 
+// GetDefaultServerWithConfigAndOpts returns a default server object that uses provided configuration
+func GetDefaultServerWithConfigAndOpts(config Config, opts ...DefaultServerOption) Server {
+	server := &DefaultServerImpl{
+		config: config,
+	}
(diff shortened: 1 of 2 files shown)


orion/core.go
```diff
@@ -513,6 +513,17 @@ func GetDefaultServer(name string, opts ...DefaultServerOption) Server {
 	return server
 }
 
+// GetDefaultServerWithConfigAndOpts returns a default server object that uses provided configuration
+func GetDefaultServerWithConfigAndOpts(config Config, opts ...DefaultServerOption) Server {
```
praveensanap, line 517: There is a function present that accept options.
if you pass the server name, it should be able to load the config and use the passed opts.

GetDefaultServer(name string, opts ...DefaultServerOption) Server 

This doesn't addressed the issue?

bhadreswar-ghuku: Config for ReadTimeout and WriteTimeout was not set. Updated.

praveensanap (approved):

PR #231, opened 2024-11-19
upgrade go, security

praveensanap:


go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/carousell/Orion
 
-go 1.17
+go 1.22
```
bhadreswar-ghuku, line 3: coverage seems reduced even when there is no change on the code. could you check

praveensanap: Something seems to be off with this tool.
https://app.codecov.io/gh/carousell/Orion/pull/231/indirect-changes
In the page it mentioned 15 indirect file coverage changes, but the change% is missing.

bhadreswar-ghuku (approved):

